PR #111, opened 2014-04-13
Add tests (WIP)

thomasf:
This is a work in progress


thomasf:
Closing this, will probably never have time to do it properly anyway (because so many modes and combinations)

PR #145, opened 2014-12-07
Updated theme for next release of magit

andschwa:
This updates the Solarized theme for the 'next' branch of magit. I
prioritized adhering to magit's current color usage, overriding with the
respective Solarized colors. Second priority was keeping the look of the
current version of magit, especially boxes around references in the
log. All magit faces which defined a background/foreground are defined
here with the respective Solarized adaptive color, and the vast majority
of faces that inherit from others are not overridden.

Since the 'next' branch has not yet been released, this PR should not be
merged until that time, as it removes the support for the current
version of magit to avoid conflicts. However, if someone wants to take
the time to do the comparison, it should be possible to keep current
support for all non-conflicting faces.


andschwa:
Hey @tarsius, I hadn't realized this was in the official [roadmap](https://github.com/magit/magit/issues/1645) for Magit 2.1.0. I'm sorry for the jumping the gun! We'll come back to it at the end of Milestone 1. At least with git we can share our experiments :smile:

andschwa:
@thomasf I went through and got rid of boxes and reduced bright color usage a bit. Does Solarized have a suggested accent color usage guide?

thomasf:
@andschwa  Sadly not.  I've been meaning to write such a document since I settled on some ways to use colors and have tried to implement it across everything.
 a lot on it there are some patterns of usage that are not very outspoken.. I've been meaning to put it it writing for some years now. I'm under a  crazy stupid amount of work since a few months. 
I will create a ticket for this now with the rules I can think of right away.

thomasf:
@andschwa  I've started by just typing out some of it in a slightly messy way in the issue linked below.. It's not very good but it is a start.

https://github.com/bbatsov/solarized-emacs/issues/162

tarsius:
@andschwa No worries, it might not have been mentioned there yet at the time you proposed this.

PR #156, opened 2015-02-20
Make company-mode annotations stand out more

ane:
company-mode recently added annotation support for some meta-level information, e.g. function parameters, here shown with company-go as a backend.

2015-02-20-122910_666x208_scrot

The default face is a red foreground. This is unreadable in the dark scheme and clashes with the colour scheme.

This PR modifies it to a clear yellow which looks nice on both colour schemes.

2015-02-20-150450_652x199_scrot

I tried other colours (green, magenta, violet) that would stand out more but found that the yellow one worked  best.


thomasf:
If there arent color collisions or something similar, one of the colors should probably be ,base1 if it's on ,base02

thomasf:
As a reference this is how popup.el looks with auto-complete and go.

screen shot 2015-02-20 at 17 28 58

ane:
You're right, I changed it to base1. It should be a lot more smoother now.

thomasf:
great thanks

PR #157, opened 2015-02-21
tweak navi2ch face

myuhe:


bbatsov:
Screenshots?

thomasf:
It makes general sense to not use base03 as a fg color for headers.

PR #159, opened 2015-03-07
Fix widget-field color

andschwa:
The default color of widget fields (like in customize) was unreadable
with Solarized. This background color seems reasonably close to the
default Emacs look of a contrasting background, no style.

I'll admit that a box with style "pressed-button" looks pretty good too.


bbatsov:
Can you post a screenshot of the state before and after the change?

andschwa:
Before
before

After
after

bbatsov:
A clear improvement! :+1:

andschwa:
Glad you agree :smile:

PR #160, opened 2015-03-07
Add support for smart-mode-line

andschwa:
Note that smart-mode-line's theme should be disabled when using an
external theme like Solarized.

[smart-mode-line](https://github.com/Bruce-Connor/smart-mode-line)


bbatsov:
:+1:

thomasf:
I think the red and blue gives it too much weight

thomasf:
Maybe try to fit everything by just using the baseXX colors?

Issue #163, opened 2015-03-08
Fix smart-mode-line display


thomasf:
Current issues:

* Too many colors are used in vain.
* All faces shoud maybe be defined, depends on how this issue is resolved.
* Current defs are not compatible with solarized-high-contrast-mode-line, this is maybe not a strict requirement but might be doable without too much effort.

thomasf:
My initial thoughts are leaning towards only letting `sml/outside-modified` and `sml/sudo` stand out from everything else since those can produce the most severe side effects is missed and would not be lit most of the time during normal usage anyway.

thomasf:
There is also the issue of inactive/active mode-line facedefs not being used unless not specifying colors at all. 
@andschwa  Maybe this theme could require a an optional defcustom and  default to be turned off?  I don't feel all these colors match well to general feel of this theme..

thomasf:
I just tinkered  a bit and this might be my suggestion for a default set of colors:

```elisp
     ;; just stand out slightly more on inactive mode lines
     `(sml/filename ((,class (:foreground ,base1 :weight bold))))
     `(sml/prefix ((,class (:foreground ,base0))))
     ;; warning
     `(sml/sudo ((,class (:foreground ,orange :weight bold))))
     ;; indication of change
     `(sml/modified ((,class (:foreground ,yellow))))
     `(sml/vc-edited ((,class (:foreground ,yellow))))
     `(sml/outside-modified ((,class (:foreground ,yellow))))
```

andschwa:
Woops, sorry for not noticing this issue sooner, for some reason I wasn't watching the repo! I've let the colors sink in myself for a while now and definitely feel they can be improved. Thanks for writing up the color usage guideline by the way in #162.

My initial thoughts on things that *should* have colors:

* Running processes should remain magenta, as they are a temporary thing which needs to be highlighted
* Read-only and sudo should have colors as their indicators are *small*
* Charging/discharging should be either green/red or green/orange
* `outside-modified` should be a different color, as it conveys totally different, but important, information from modified internally

I'm testing out yours with the addition of
```
     `(sml/git ((,class (:foreground ,base0))))
```
So that a prefix even in a Git repo remains the same color.

I would *like* to highlight `sml/vc` as the current branch/tag in a repo is probably the most important information in the modeline. It's yellow right now, which means that the modified color isn't actually a change. Perhaps blue for unmodified and yellow for modified and vc-edited?

thomasf:
[...]

PR #165, opened 2015-03-17
Update smart-mode-line colors

andschwa:
Per the color conventions.

@thomasf is that sort of what you were thinking in #163?

The faces that weren't defined were those that used the default; but if the convention is to define them anyway, I'll add a commit here shortly.


andschwa:
More questions:

* What color would you recommend for a process other than magenta?
* Should inheritance of built-in be relied on or not? There are faces that could certainly just inherit default/success/warning/error.
* Is the preference to group faces logically or in the same order that they're defined in the package?

a3ammar:
Can you at least revert c4db87e8d2972b8b8111565215e108dbb799e302 until this gets merged? Surely too many colors used in vain is better than no colors at all for the dark theme.

andschwa:
I would agree @a3ammar and did so in my own fork, but sadly here it's not up to me :/

andschwa:
I've gone ahead and rebased onto master so this can merge cleanly. We may still want to tweak more colors.

Malabarba:
Once this is done, if someone would like to make a couple of small-ish screenshots I'd be very happy to add them to sml's frontpage. I'm afraid the current screenshots have become outdated.

thomasf:
@a3ammar

My opinion is that the current look fits better into the general look of this
theme than it does if more colors are added. This is in part because of
"issues" uncommon in other palettes, solarized does not have bright/dark
version of "accent colors" which makes balancing many colors at once not work
as well.. I've spent many many hours figuring out how to get a good overall
balance over the years and it took at least a year or so until I reached the
decision to systematically reduce overall color usage..

Btw, some colors are currently defined for sml:

* Plain foreground text for the bulk of text: Most individual indicators has
  their specific locations there is very little need to highlight the locations
  further. Also, humans are generally adept at relative locations in (in space)
  so if the mode line itself is easy to spot the rest kind of follows naturally.

* The buffer name is using bold base1, emphasized fg color. (The single most
  important information on the mode line)

* sudo is orange (as a warning to be careful)

* modified/read-only/outside-modified are cyan so they are (as a group) clearly
  visible to the user.

* vc-edited is green.. IIRC, someone said this specific indicator was super
  important to be highlighted so I added the color least intrusive on any
  backgroun. The default location of this indicator is also around the center
  of the mode line which is good because it does not disturb the overall
  left/right weight.

thomasf:
And there is another popular solarized theme for emacs as well.. Having two or three actively maintained  emacs solarized themes which also implements exactly the same looks would also be a waste of everyone's time..

a3ammar:
The problem is the dark version defaults to the light version colors. This is what it looks like on my system:
screen shot 2015-04-14 at 11 32 31 am

thomasf:
@a3ammar thats not right., have you done the `(setq sml/theme nil)` before sml is loaded ?

thomasf:
Here;s how it should look currently:

image

image

thomasf:
closing this because this theme has an aggressive anti color pasta ideology :)

PR #170, opened 2015-03-26
Add neotree support

ryanartecona:
Add support for [neotree](https://github.com/jaypei/emacs-neotree). Colors were taken from the Vim scheme for NERDTree.

#### Before:
screen shot 2015-03-25 at 11 04 52 pm

#### After:
screen shot 2015-03-25 at 11 04 05 pm

### DIFF
--- a/solarized.el
+++ b/solarized.el
@@ -1460,6 +1460,13 @@ customize the resulting theme."
      `(navi2ch-message-link-face ((,class (:weight bold))))
      `(navi2ch-message-url-face ((,class (:weight bold))))
      `(navi2ch-message-citation-face ((,class (:foreground ,magenta))))
+;;;;; neotree
+     `(neo-banner-face ((,class (:foreground ,base01))))
+     `(neo-header-face ((,class (:foreground ,blue))))
+     `(neo-root-dir-face ((,class (:foreground ,base1 :weight bold))))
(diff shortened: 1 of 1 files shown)


thomasf:
Change the expander to use base01 or base1 or similar, otherwise great.

thomasf:
direx for reference 

image

thomasf:
I would probably drop the magenta and green at the top and use maybe bold normal or emphasized color there as well. I'm not familiar with neotree so I'm not aware if those are being used for other things as well.

ryanartecona:
Done! Updated screenshot:
screen shot 2015-03-27 at 12 21 25 am

The cyan line behaves like the `../` relative directory, so it should probably be colored as a directory. The green line is effectively `$PWD` within neotree, so while that one will *also* always be a directory, I think it's more important that it's highlighted from its surroundings (a bunch of other directories, typically). Every single face that neotree uses should be shown in the updated screenshot.

I think the `$PWD` line could stand to have its own color (green or cyan?), but TBH I care about this far less than at least the plugin using all solarized colors.

thomasf:
yeah, things like ,base1 usually stand out enough not to warrant adding additional colors for titles and such.. Looks good.

thomasf:
The combination of blue/green/cyan is good and those are the dominating colors used in this theme. I have no defined strategy for choosing between cyan or green.. It depends on the context.

PR #173, opened 2015-03-29
Add cider-repl-err-output-face

cichli:
Generally the faces for `cider-repl-mode` seem fine, but this particular one inherits from `font-lock-warning-face` which is underlined in the theme. One alternative is to remove the underline from that face entirely, or add an option to use underline less. FWIW I've only ever found it noisy/distracting. Since the theme uses red/orange pretty sparingly, anything in that colour tends to stand out anyway without being underlined.

Before:
screenshot 2015-03-29 22 03 32

After:
screenshot 2015-03-29 22 04 51


thomasf:
You might be right about removing underline from the warning face..

PR #178, opened 2015-04-19
Allow users to disable the overriding of the OS mouse cursor.

jonathanj:
The theme casually overrides the OS mouse cursor face and I can't figure out how to reverse these settings once they've been made. (One can get close but it's not the same.)

This PR adds a customization variable (that defaults to `t`) to enable this behaviour.

### DIFF
--- a/solarized.el
+++ b/solarized.el
@@ -109,6 +109,11 @@ Also affects `linum-mode' background."
   :type 'boolean
   :group 'solarized)
 
+(defcustom solarized-override-mouse-pointer t
+  "Whether the color of the mouse pointer is changed according to the theme."
+  :type 'boolean
+  :group 'solarized)
+
 ;;; Utilities
 
 (defun solarized-color-name-to-rgb (color &optional frame)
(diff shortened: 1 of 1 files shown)


thomasf:
For wich windowing toolkits does that happen?

jonathanj:
@thomasf OS X in this case.

thomasf:
Ait, I'll go find my laptop then... Is it happening for the version on http://emacsformacosx.com/ ? It just might be that I don't use the mouse enough to even notice it...

jonathanj:
I'm using the railwaycat binary via a Homebrew tap. <https://github.com/railwaycat/emacs-mac-port>

jonathanj:
I originally reports the issue to the spacemacs repo and included screenshots: https://github.com/syl20bnr/spacemacs/issues/1181

thomasf:
I'm installing that to take a look, seems to take a while...
Just to be certain, is the actual mouse pointer icon colored by those defs?

jonathanj:
I'm not 100% sure about defs other than `mouse` (since I have no idea where the others apply) but changing `mouse` to use `red` definitely changes the mouse cursor.

Perhaps only conditionally setting the primary cursor might be a better behaviour choice?

thomasf:
I'm thinking that it should not be changed at all..
The gtk version (which is my primary emacs setup) does not support it but I guess at least motif/lucid builds does. The faces affecting look'n'feel outside of the emacs frame are a bit randomly supported over the broader spectrum.

However, changing the mouse cursor should signal something, just randomly changing isn't really useful.. The user can pick a good os level mouse cursor color of choice..

jonathanj:
Obviously that would suit me. Perhaps then overriding it should default to `nil` for now (so if someone wants the behaviour they can turn it back on) with the aim to remove it altogether after some feedback?

thomasf:
Nah, better to just remove it.
`popup-menu-mouse-face` is probably something else,  It seems to color the items when being pointed at.

jonathanj:
Okay, so should I just remove the `mouse` def and leave the others unchanged?

thomasf:
I actually noticed a couple of errors there.. Will push the change is a minute or so.

thomasf:
The blue here was the popup menu one, the other ones were also related to that package.
image

thomasf:
Closing this issue for now, should be solved now.

jonathanj:
Thanks for the swift help, Thomas!

PR #184, opened 2015-04-25
make fontsize for some Org agenda elements behave in accordance with solarized-scale-org-headlines

holtzermann17:
Items in the Org Mode Agenda and TODO list were too small to be readable on my monitor.  I've disabled font scaling for those items, following the value set by `solarized-scale-org-headlines`.  Test with `C-a a` or `C-a t`.  

It's possible that you'll want to use another variable name to control these settings, but this behaviour is better than the current default in which the font is tiny on a high-res monitor.

### DIFF
--- a/solarized.el
+++ b/solarized.el
@@ -1483,7 +1483,8 @@ customize the resulting theme."
 ;;;;; org-mode
      `(org-agenda-structure
        ((,class (:foreground ,base1 :background ,base02
-                             :weight bold :slant normal :inverse-video nil :height ,solarized-height-plus-1
+                             :weight bold :slant normal :inverse-video nil 
+                             ,@(when solarized-scale-org-headlines :height ,solarized-height-plus-1)
                              :underline nil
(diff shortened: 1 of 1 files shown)


thomasf:
Hmm if `:height 1.x` is smaller than default text the reason is probably a font configuration issue on your system.. What OS are you using?

tarsius:
Since after a year op has not replied to your suggestion, I think you can close this now.

holtzermann17:
@tarsius I guess I'm OP?  I think it might make sense for someone to *test* the suggestion.  I didn't see the earlier message until now.  Sorry!  Anyway as chance would have it I saw the message today and I can confirm that the earlier suggestion from @thomasf **does indeed** work, with the latest version of solarized.el and latest theme files.

@thomasf For the record I'm on Ubuntu 14.04, but running it on a Macbook Pro w/ a Retina display and the latter is likely the cause of the old scaling issues.

tarsius:
Ah yes, should probably have suggested that instead :-)

Issue #180, opened 2015-04-22
Look at possibly introducing an alternative base0'ish color to separate certain regions.


thomasf:
## The problem

Sometimes an alternative background color to differentiate regions is needed.
The go to face for this is `base02` but that has issues. `base02` cannot be
assumed to work since there is not always controllable that `base01` won't be
put on top of. `base02`/`base01` have too little contrast between eachother for
that to work. `hl-line-mode` users might also get in "trouble" if larger static
blocks have `base02` as background colors since that's what `hl-line-mode` also
uses to show the current line.

For most cases it  has worked good enough to just use `base00` instead but in some
instances it would be beneficial to have an alternative.

## The new color

**I have not experimented at all yet but I'm thinking that a tad more greyish version of `base03` should be evaluated.. It's not an easy though since thats basically already is what `base02` is all about.

## examples of potential uses

There are some exceptional circumstances where this color would make sense.  Primarily it would be used when one mode/language is embedded into the flow of another where..  org-blocks, `THIS THING` and so on... IIRC some table generating modes would also like to have highlighted column/row backgrounds in some situations.

### org (babel) blocks

An example of something more extreme than what I'm thinking of:

![](http://orgmode.org/worg/org-contrib/babel/images/fontified-src-code-blocks.png)

Current solarized look:

![image](https://cloud.githubusercontent.com/assets/330847/7270084/a1cee542-e8d9-11e4-8fab-0c776274e433.png)

The usage of the `base01` fg color  for  `org-block` is suboptimal but all other things I tried felt like too much or too little. With regular `base0` it was really hard to read docs with many small example/source blocks .

### sx.el

Upstream screehshot:

![](https://github.com/vermiculus/sx.el/blob/master/list-and-question.png?raw=true)

My work in progress theming:

![image](https://cloud.githubusercontent.com/assets/330847/7250104/aed05f92-e81f-11e4-8272-05212ffc375f.png)

thomasf:
this might be something:

![image](https://cloud.githubusercontent.com/assets/330847/7667470/aa6a8d44-fc09-11e4-9336-d4aa036b63c4.png)
![image](https://cloud.githubusercontent.com/assets/330847/7667478/cb33fcd6-fc09-11e4-8af1-7f91406015f5.png)

thomasf:
In the end I decided against this, there isn't enough contrast margin between the basic solarized contrasts for this to work,.

[...]

PR #186, opened 2015-04-29
Added `solarized-org-block-highlight` switch.

szozda:
If not `nil`, it causes the org blocks to have highlighted background
and standing out beginning and ending lines. 

Screenshot of `solarized-light` with `solarized-org-block-highlight` set to `t`.
screenshot-07

Screenshot of `solarized-dark` with `solarized-org-block-highlight` set to `t`.
screenshot-06

### DIFF
--- a/solarized.el
+++ b/solarized.el
@@ -109,6 +109,11 @@ Also affects `linum-mode' background."
   :type 'boolean
   :group 'solarized)
 
+(defcustom solarized-org-block-highlight nil
+  "Whether to highlight org blocks."
+  :type 'boolean
+  :group 'solarized)
+
 ;;; Utilities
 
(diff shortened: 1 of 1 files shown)


thomasf:
Thanks, I might wait until #180 is decided since it's kind of related.

szozda:
Thanks. With the current colors it actually stands out a bit too much (especially the beginning and ending lines) so that is a good idea.

thomasf:
@szozda I think it might  be going in this direction https://github.com/bbatsov/solarized-emacs/issues/180#issuecomment-102664347

szozda:
Looks good. Thanks for the update.

thomasf:
In the end I decided against this, there isn't enough contrast margin between the basic solarized contrasts for this to work,.

PR #189, opened 2015-06-27
Support version 2.1.0 of Magit

tarsius:
This should be merged after I have released version `2.1.0` which will happen on the 1th of July, but not earlier.

This is not quite ready yet. Most importantly the dark variant is completely missing right now. I hope to get to that before the release, but cannot promise anything. And the light variant also still needs some fine-tuning.

### DIFF
--- a/solarized.el
+++ b/solarized.el
@@ -610,18 +610,27 @@ customize the resulting theme."
      `(custom-button-unraised ((,class (:inherit underline))))
      `(custom-button-pressed-unraised ((,class (:inherit custom-button-unraised :foreground ,magenta))))
 ;;;;; diff
-     `(diff-added ((,class (:foreground ,green :background nil))))
-     `(diff-changed ((,class (:foreground ,blue :background nil))))
-     `(diff-removed ((,class (:foreground ,red :background nil))))
-     `(diff-header ((,class (:background ,base03))))
+     `(diff-added   ((,class (:foreground ,green))))
+     `(diff-changed ((,class (:foreground ,blue))))
+     `(diff-removed ((,class (:foreground ,red))))
+     `(diff-refine-added
+       ((((class color) (background light))
+         (:background ,(solarized-color-blend "#ddffdd" green 0.7)))
+        (((class color) (background dark))
+         (:background ,(solarized-color-blend "#446644" green 0.7)))))
+     `(diff-refine-changed
+       ((((class color) (background light))
+         (:background ,(solarized-color-blend "#ddddff" blue 0.7)))
+        (((class color) (background dark))
+         (:background ,(solarized-color-blend "#444466" blue 0.7)))))
+     `(diff-refine-removed
+       ((((class color) (background light))
+         (:background ,(solarized-color-blend "#ffdddd" red 0.7)))
+        (((class color) (background dark))
+         (:background ,(solarized-color-blend "#664444" red 0.7)))))
+     `(diff-header  ((,class (:background ,base03))))
      `(diff-file-header
        ((,class (:background ,base03 :foreground ,base0 :weight bold))))
-     `(diff-refine-added ((,class (:foreground ,green :background ,base03
-                                               :inverse-video t))))
-     `(diff-refine-change ((,class (:foreground ,blue :background ,base03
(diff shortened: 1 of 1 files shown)


thomasf:
Thats on my birthday, great present :)

tarsius:
I have no pushed what might very well be the last revision before the magit release. Maybe I will run into some problems during the next 24 hours or so and then address it. But probably not.

One last change I had to make was to adjust the diff refinement faces when using the dark variant. All the other faces work for the dark variant. Mostly by accident, resp. because of the overall consistency of the solarized themes. Speaking of which; the diff related faces (and some others too) do use colors that are not part of the solarized palette. This is necessary because these faces should be very similar but not quite.

There is certainly room for improvement, but this should work for now.

I will ping you once I have tagged the release.

thomasf:
I did some minor changes when I tested it.. I replaced orange with basexx's and made tags cyan.. I do not have time to tweak those so I'll ignore those for now. 
And yeah.. the colors would be a little easier to work with using other methods than RGB blending.. the CIE-LAB color functions has been in color.el since emacs 24.(something) so we could probably start assume that people has that at this point,..

thomasf:
This is my mindor adjustments form yesterday.,. not sure about all of them. Did not have time to start scratching on the diff/log stuff..

```diff
diff --git a/solarized.el b/solarized.el
index 42f5996..40c935b 100644
--- a/solarized.el
+++ b/solarized.el
@@ -1284,13 +1284,13 @@ customize the resulting theme."
             :foreground ,orange
             :weight bold))))
      `(magit-diff-lines-heading          ((t (:background ,orange
-                                              :foreground ,base3))))
+                                              :foreground ,base0))))
      `(magit-diff-context-highlight      ((t (:background ,base02))))
      `(magit-diffstat-added              ((t (:foreground ,green))))
      `(magit-diffstat-removed            ((t (:foreground ,red))))
 ;;;;;; popup
      `(magit-popup-heading             ((t (:foreground ,yellow  :weight bold))))
-     `(magit-popup-key                 ((t (:foreground ,green-d :weight bold))))
+     `(magit-popup-key                 ((t (:foreground ,green   :weight bold))))
      `(magit-popup-argument            ((t (:foreground ,green   :weight bold))))
      `(magit-popup-disabled-argument   ((t (:foreground ,base01  :weight normal))))
      `(magit-popup-option-value        ((t (:foreground ,blue    :weight bold))))
@@ -1298,8 +1298,8 @@ customize the resulting theme."
      `(magit-process-ok    ((t (:foreground ,green :weight bold))))
      `(magit-process-ng    ((t (:foreground ,red   :weight bold))))
 ;;;;;; log
-     `(magit-log-author    ((t (:foreground ,orange))))
-     `(magit-log-date      ((t (:foreground ,base0))))
+     `(magit-log-author    ((t (:foreground ,base0))))
+     `(magit-log-date      ((t (:foreground ,base01))))
      `(magit-log-graph     ((t (:foreground ,base1))))
 ;;;;;; sequence
      `(magit-sequence-pick ((t (:foreground ,yellow-d))))
@@ -1322,7 +1322,7 @@ customize the resulting theme."
 ;;;;;; references etc.
      `(magit-dimmed         ((t (:foreground ,base01))))
      `(magit-hash           ((t (:foreground ,base01))))
-     `(magit-tag            ((t (:foreground ,orange :weight bold))))
+     `(magit-tag            ((t (:foreground ,cyan   :weight bold))))
      `(magit-branch-remote  ((t (:foreground ,green  :weight bold))))
      `(magit-branch-local   ((t (:foreground ,blue   :weight bold))))
      `(magit-branch-current ((t (:foreground ,blue   :weight bold :box t))))
```

tarsius:
I have pushed the release tag now, so you should merge this now.

Please leave out the first hunk of your patch for now. I didn't have the time to test, but am pretty sure that would have drawbacks.

I have used a different color for keys and arguments because in the optimal case these are the same - but not always! I think using different colors helps recognizing this. But either way is fine with me.

The other changes are harmless, but I think I still prefer my colors. But doesn't really matter much.

bbatsov:
We'll go with your colours for now. Thanks!

PR #193, opened 2015-07-14
Add coloring for highlight-numbers

spirani:
Hi, this pull request sets the color for the "highlight-numbers" package (https://github.com/Fanael/highlight-numbers), which highlights numbers in source code. I set it to the solarized "violet" color.

Screen shot for the dark theme: 
example

Screen shot for the light theme:
example2

### DIFF
--- a/solarized.el
+++ b/solarized.el
@@ -1079,6 +1079,8 @@ customize the resulting theme."
 ;;;;; highlight-indentation
      `(highlight-indentation-face ((,class (:background ,base02))))
      `(highlight-indentation-current-column-face((,class (:background ,base02))))
+;;;;; highlight-numbers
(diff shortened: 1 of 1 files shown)


thomasf:
violet is probably good since it's almost unused. 
Personally I would advice against using it, this theme actually removed colors in several cases to create a slightly flatter overall look.

Why `:bold nil` ?   `:bold unspecified` might also work, have you tried that?

spirani:
Alright, thanks for the tip.

I used ```:bold nil``` because by default, highlight-numbers inherits its face from font-lock-constant-face, which solarized sets with a bold font. Leaving out ```:bold nil``` or using ```:bold unspecified``` instead keeps the bold font for highlighting numbers, which I think is distracting.

thomasf:
ok, merging then.

Issue #196, opened 2015-07-22
child theme use


blaenk:
I was very excited when I saw that child themes were possible in this. The ability to reuse the solarized color variables and layer customizations on top of the theme is amazing, since you don't have to hack on and fork the theme itself.

I can't seem to get it to work though. I am able to `load-theme` it, but my customizations aren't showing up. I emit a `(message "test")` to make sure the theme function is being called, it is.

I have this in `init.el`:

``` elisp
(defun solarized-blaenk-theme ()
  "My solarized child theme."
  (message "written")
  (custom-theme-set-faces
   theme-name
   `(default ((,class (:foreground ,red :background ,base03))))
   `(rainbow-blocks-depth-1-face ((,class (:foreground ,red))))
   ))

(provide 'solarized-blaenk)
(load-theme 'solarized-blaenk-light t)
```

This is what I have in ~/.emacs.d/themes which is in my custom theme load path:

``` elisp
(require 'solarized)
(require 'solarized-blaenk)

;; This files needs to be places iside the custom-theme-load-path list
(deftheme solarized-blaenk-light "The light variant of the Solarized colour theme")
(create-solarized-theme 'light 'solarized-blaenk-light 'solarized-blaenk-theme)

(provide-theme 'solarized-blaenk-light)
```

I'm not sure if that's how you're supposed to split it up; I'm new to emacs.

As is, solarized light does load but the customizations aren't showing up. The interesting (?) thing is that if I go to the theme file which includes the `deftheme` call and I `M-x eval-buffer`, my customizations _do_ show up as I would expect (i.e. the default font face shows up red).

What am I doing wrong? I really really want to get this to work, it sounds like an awesome feature :)

blaenk:
Perhaps `theme-name` in `solarized-blaenk-theme` should be `'solarized-blaenk-light`? I just copied what was in the child theme example. I changed it though and that had no effect.

blaenk:
It works if I add this line after my `load-theme` line:

``` elisp
(create-solarized-theme 'light 'solarized-blaenk-light 'solarized-blaenk-theme)
```

Perhaps it's something to do with load order?

blaenk:
After playing around with it I noticed that consecutive calls to `custom-theme-set-faces` don't take effect. I think that may be the root of the issue.

blaenk:
Trying to figure things out in this [stackexchange thread](http://emacs.stackexchange.com/questions/14205/redefine-theme-spec).

[...]

PR #198, opened 2015-07-24
introduce solarized-with-color-variables macro

### LINKED ISSUE #197: with-color-variables macro
Would be cool if there were a macro like you have in the zenburn theme package that exposes all of the color variables, instead of keeping it contained in a let binding form.

blaenk:
This is similar to the macro available in the zenburn theme. This
provides a nice way to reuse the solarized color variables which makes
customizations possible, since child themes don't appear to be working.

See #196

Closes #197

### DIFF
--- a/solarized.el
+++ b/solarized.el
@@ -153,14 +153,10 @@ Alpha should be a float between 0 and 1."
                     (solarized-color-name-to-rgb color2))))
 
 ;;; Setup Start
-
-(defun create-solarized-theme (variant theme-name &optional childtheme)
-  "Create a VARIANT of the theme named THEME-NAME.
-
-When optional argument CHILDTHEME function is supplied it's invoked to further
-customize the resulting theme."
-;;; Color palette
-  (let* ((class '((class color) (min-colors 89)))
+(defmacro solarized-with-color-variables (variant &rest body)
+  (declare (indent 0))
+  `(let* ((class '((class color) (min-colors 89)))
+	 (variant ,variant)
          (s-base03    "#002b36")
          (s-base02    "#073642")
          ;; emphasized content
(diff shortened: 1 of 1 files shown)


blaenk:
Also the xterm variables were outside of the custom-theme-set-variables call so I brought them into it.

thomasf:
Great!

cdlm:
@blaenk how do you use this exactly to customize solarized-dark or solarized-light? define a complete set of faces using colors from the macro?

My use-case is I'm pretty happy with both solarized themes, but I'd like headings (in org, markdown, latex…) to use colors in a more spectrum-like sequence (orange, yellow, green, cyan, blue, violet). I'd also like the definitions to be a proper theme, so that they stay along after switching themes. Child themes seemed nice, but since they don't work, the only solution I see is basically a fork: copy-pasting `create-solarized-theme` into my own theme and tweak color choices there…

blaenk:
This macro basically exposes solarized's color variables within its body. So for example, you can then use `custom-theme-set-faces 'some-theme` to add customizations to an existing theme (e.g. solarized-light) based off of those colors. Below is an example where I do so to set the background color of the `pulse-highlight-start-face` to the value of the solarized color variable `base02` for the light variant.

This is a rough, untested example.

``` elisp
(solarized-with-color-variables 'light
 (custom-theme-set-faces
  'solarized-light

  ;; pulse highlight
  `(pulse-highlight-start-face ((,class (:background ,base02)))))
```

cdlm:
Well sure, I got that far… the problem is, as said somewhere else, that the second call to `custom-theme-set-faces` has no effect. I guess you could extend a theme this way, but I haven't yet managed to override color choices from the base theme…

blaenk:
You can take a look at my configuration, where I created a separate theme and layer it over solarized.

thomasf:
It could maybe be added to the readme or docstring as an example?

PR #200, opened 2015-08-04
Add file child-theme-example/.nosearch

tarsius:
This prevents that the directory "child-theme-example/" is added to the `load-path`.


thomasf:
Probably a good idea, thanks.

PR #205, opened 2015-09-04
Use cl-lib instead of cl

tarsius:


solarized-theme-utils.el
```diff
@@ -24,7 +24,7 @@
 ;;
 
 ;;;; Code:
-(require 'cl)
+(require 'cl-lib)
```
bbatsov: We also need to add a dep to `cl-lib` from GNU ELPA for older Emacsen (e.g. 24.1).

tarsius:
Ah sorry, did not realize that was distributed with the package. I've added the dependency now.

thomasf:
It's actually not really  meant to be distributed in the package. . 
Thanks!

syl20bnr:
Sorry for hi-jacking this closed PR but I'm curious about this replacement, why `cl-lib` is better than `cl` ? is it just a matter of namespace ?

thomasf:
Yes, cl.el was removed from emacs and replaced by cl-lib a few releases ago.. cl.el is now deprecated and just aliases cl-lib.

thomasf:
To keep compatibility with earlier emcs versions cl-lib is also published to elpa so that emacses which does not have it can download it.. That's the main reason to have it as a package depencency.

syl20bnr:
Thank you!

Le mercredi 16 septembre 2015, Thomas Frössman <notifications@github.com> a
écrit :

> To keep compatibility with earlier emcs versions cl-lib is also published
> to elpa so that emacses which does not have it can download it.. That's the
> main reason to have it as a package depencency.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/bbatsov/solarized-emacs/pull/205#issuecomment-140746120>
> .
>


-- 
-syl20bnr-

PR #206, opened 2015-09-08
Faces for avy-goto-*

fleimgruber:
These faces should be close to isearch ones. The default ones are color pasta al dente when used together with solarized.


bbatsov:
:+1:

PR #207, opened 2015-09-13
Fix typo in child theme examples

PythonNut:


bbatsov:
:+1: Thanks!

thomasf:
Thanks. btw. child themes are probably broken ( #196 ) but could possibly be updated to work somehow because of color palette generation was recently moved to it's own function ( #198 )

PythonNut:
@thomasf Sadly, I only figured out that child themes were broken after trying them out myself. I was really excited about them too: they look like a very clean way to personalize the face definitions. (Right now, my face definitions are scattered around in various hooks and `with-eval-after-load` blocks and honestly, it doesn't feel very clean)

thomasf:
It's probably fixable, it just stopped working because of changes in one of the 24.x emacs versions.. I just have not had time to do it.. I used to use it all the time to evaluate changes to the theme before pushing to this repository. I'll look into it when I get a chance.

PR #208, opened 2015-10-01
Add a definition for helm-header-line-left-margin

PythonNut:
This should help people who use the setting `(setq helm-echo-input-in-header-line t)`.


thomasf:
thanks

PR #209, opened 2015-10-21
Do not depend on the f and s packages for tests

tarsius:
```
`s' was not used directly but is a dependency of `f'.  The only
function from `f' which was used here is `f-expand' which is
defined like so:

  (defun f-expand (path &optional dir)
    "Expand PATH relative to DIR (or `default-directory')."
    (directory-file-name (expand-file-name path dir)))

Since the `directory-file-name' part is irrelevant in our usage,
we might as well just use `expand-file-name'.

Also remove some empty lines and the empty and not properly
terminated "Commentary" section.
```

### DIFF
--- a/tests/emacs-visual-test.el
+++ b/tests/emacs-visual-test.el
@@ -1,27 +1,22 @@
 ;;; emacs-visual-test.el ---
 
-;;; Commentary:
-;;
-
 (load-file "init.el")
-(require 's)
 (require 'dash)
-(require 'f)
 (setq header-line-format "header line content")
 
 (defconst screenshots-directory
-  (f-expand "screenshots" tests-directory))
+  (expand-file-name "screenshots" tests-directory))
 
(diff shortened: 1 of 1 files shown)


thomasf:
Sure, we don't really need either of them.

To me `s` brings a lot of string manipulation stuff together in a nice way under a common prefix while `f` actually changes too much from the standard emacs functions to be really useful.  

I think that spending some time on making this work is the way forward for the tests project (?):  https://github.com/bbatsov/solarized-emacs/tree/add-tests/features I can't remember why but I could not get it to work as I wanted within a few hours and I have not revisited that project since.

tarsius:
I cannot really help you with that, sorry. I was just cleaning out some old feature branches from installed packages and came across this one from the beginning of the year. Instead of just throwing it away I thought I might as well pimp it a little and open a pr :-)

PR #210, opened 2015-10-21
Add support for dired+

### LINKED ISSUE #166: Support for dired+
First of all, thanks for this wonderful theme. I really love your latest work with a better support for ```smart-mode-line```.

Do you think it will be possible to support ```dired+``` as well? Colors for ```dired+``` are pretty ugly compared to what solarized-theme offers, say, for ```org-mode```. I set ```font-lock-maximum-decoration``` to default for ```dired-mode```, basically reverting back to default dired colors and losing all ```dired+``` highlightings.

I know this is not an issue per se, just asking whether you are planning to support ```dired+```.

mikolajb:
Dired+ faces, fixes #166


thomasf:
By just looking at it I think it might use slightly too many colors , see https://github.com/bbatsov/solarized-emacs/blob/master/DEV-GUIDE.md

I'll check it out later and make a decision.

bbatsov:
Any progress here?

mikolajb:
Maybe it would help

solarized

thomasf:
referring to https://github.com/bbatsov/solarized-emacs/blob/master/DEV-GUIDE.md I primarily object to the permissions color pasta and the use of red and magenta in there as well. I probably recommend the normal foreground color for all of it.

thomasf:
I'm closing this issue because no reply

PR #227, opened 2016-04-06
Fix rst-mode headers

### LINKED ISSUE #225: broken rst-mode headers would result in disco lights
the `rst-mode` faces use the wrong names (`rst-level-1-face` instead of `rst-level-1`) so they aren't used at all. which happens to be a good thing since it will result in disco light effects:

```elisp
;;;;; rst-mode
     `(rst-level-1 ((,class (:background ,yellow  :foreground ,base03))))
     `(rst-level-2 ((,class (:background ,cyan    :foreground ,base03))))
     `(rst-level-3 ((,class (:background ,blue    :foreground ,base03))))
     `(rst-level-4 ((,class (:background ,violet  :foreground ,base03))))
     `(rst-level-5 ((,class (:background ,magenta :foreground ,base03))))
     `(rst-level-6 ((,class (:background ,red     :foreground ,base03))))
```

a much better alternative would be to just reuse the `org-level-x` faces, just like the `outline-x` faces do:

```elisp
;;;;; rst-mode
     `(rst-level-1 ((,class (:inherit org-level-1))))
     `(rst-level-2 ((,class (:inherit org-level-2))))
     `(rst-level-3 ((,class (:inherit org-level-3))))
     `(rst-level-4 ((,class (:inherit org-level-4))))
     `(rst-level-5 ((,class (:inherit org-level-5))))
     `(rst-level-6 ((,class (:inherit org-level-6))))
```

wbolster:
Fixes #225.

### DIFF
--- a/solarized.el
+++ b/solarized.el
@@ -1734,10 +1734,10 @@ customize the resulting theme."
      `(rainbow-delimiters-unmatched-face
        ((,class (:foreground ,base0 :background ,base03 :inverse-video t))))
 ;;;;; rst-mode
-     `(rst-level-1 ((,class (:inherit org-level-1 :height ,solarized-height-plus-4))))
-     `(rst-level-2 ((,class (:inherit org-level-2 :height ,solarized-height-plus-3))))
-     `(rst-level-3 ((,class (:inherit org-level-3 :height ,solarized-height-plus-2))))
-     `(rst-level-4 ((,class (:inherit org-level-4 :height ,solarized-height-plus-1))))
+     `(rst-level-1 ((,class (:inherit org-level-1))))
+     `(rst-level-2 ((,class (:inherit org-level-2))))
+     `(rst-level-3 ((,class (:inherit org-level-3))))
+     `(rst-level-4 ((,class (:inherit org-level-4))))
      `(rst-level-5 ((,class (:inherit org-level-5))))
      `(rst-level-6 ((,class (:inherit org-level-6))))
 ;;;;; rpm-mode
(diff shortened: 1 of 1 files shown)


solarized.el
```diff
@@ -1734,12 +1734,12 @@ customize the resulting theme."
      `(rainbow-delimiters-unmatched-face
        ((,class (:foreground ,base0 :background ,base03 :inverse-video t))))
 ;;;;; rst-mode
-     `(rst-level-1-face ((,class (:background ,yellow   :foreground ,base03))))
-     `(rst-level-2-face ((,class (:background ,cyan    :foreground ,base03))))
-     `(rst-level-3-face ((,class (:background ,blue    :foreground ,base03))))
-     `(rst-level-4-face ((,class (:background ,violet  :foreground ,base03))))
-     `(rst-level-5-face ((,class (:background ,magenta :foreground ,base03))))
-     `(rst-level-6-face ((,class (:background ,red     :foreground ,base03))))
+     `(rst-level-1 ((,class (:inherit org-level-1 :height ,solarized-height-plus-4))))
```
bbatsov: Do we need the different heights?

wbolster:
no imho the heights can go; i put them in since they appear elsewhere, but it's bad for text editing anyway.

thomasf:
Yes, the usage of the different heights should be deiced upon for each use case. Some modes/text formats are better suited for them than others. I would probably prefer not to have them in rst-mode.

wbolster:
yeah i agree; updated this pr already.

thomasf:
thanks

PR #228, opened 2016-04-19
Add faces for minibuffer hints of git-timemachine

lunaryorn:
See https://github.com/pidu/git-timemachine for information about git-timemachine.

I picked `s-orange` and `s-yellow` for no other reason than that git-timemachine itself uses "orange" and "yellow" respectively for these faces.  I've read the devguide about `orange` but I looks good and I don't think a user could confuse this message with an error message:

bildschirmfoto 2016-04-19 um 16 41 29

However, if you'd like to use different colours—maybe there are standard colours for this kind of display—please tell me and I'll update the pull request.


thomasf:
Yeah, it seems fine. Thanks.

lunaryorn:
@thomasf Thank you very much for merging :blush:

PR #230, opened 2016-05-05
Autoload solarized-color-blend to avoid breaking customisation

purcell:
When saving customisation after enabling this theme, the saved settings for certain variables (e.g. the highlight symbol faces) get set to forms which reference solarized-color-blend. If the saved theme is later switched to a different theme, then those variables settings break, because solarized will not be loaded at the time they are evaluated.

### DIFF
--- a/solarized.el
+++ b/solarized.el
@@ -139,6 +139,7 @@ RED, GREEN, and BLUE should be numbers between 0.0 and 1.0, inclusive."
   (format "#%02x%02x%02x"
           (* red 255) (* green 255) (* blue 255)))
 
+;;;###autoload
 (defun solarized-color-blend (color1 color2 alpha)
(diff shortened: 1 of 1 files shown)


thomasf:
Thanks.. 

Maybe the situation should be avoided. I was under the impression that things were arranged so that evaluation always happens when the theme is loaded but I haven't gone into any kind of length to verify that and that's obviously not the case.

I'll merge this to avoid the breaking because I don't have time to look at whats going on atm.

purcell:
Thanks. For future reference, here's how to replicate the problem:
- Enable solarized using customize-themes, and save the customisation
- Restart Emacs
- Enable a different theme instead, and save the customisation
- Restart Emacs again

Even with this fix, you'll notice at the end of this process that there will be references to `solarized-color-blend` in the custom file.

PR #234, opened 2016-05-15
Don't duplicate code from color.el

purcell:
This theme targets Emacs 24, so it can simply directly use the color-name-to-rgb and color-rgb-to-hex, since those functions have been available in color.el since Emacs 24.1. In fact, the code removed by this commit has apparently been copied-and-pasted from there, and prefixed with "solarized-".

### DIFF
--- a/solarized-theme-pkg.el
+++ b/solarized-theme-pkg.el
@@ -2,4 +2,4 @@
   "solarized-theme"
   "1.2.2"
   "The Solarized color theme, ported to Emacs."
-  '((cl-lib "0.5") (dash "2.6.0")))
+  '((emacs "24.1") (cl-lib "0.5") (dash "2.6.0")))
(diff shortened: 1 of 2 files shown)


purcell:
I should note that the info around whether you're supporting Emacs 23 is contradictory, because the README says:

> Solarized for Emacs is tested only under Emacs 24, but should be working under Emacs 23 as well. The theme is implemented in terms of customizations and deftheme and does not require the color-theme-package.

while `solarized.el` says

> ;; Don't forget that the theme requires Emacs 24.

Might be a good opportunity to simply require Emacs 24.

thomasf:
Yeah. The reason for the copying was compatibility with older Emacs
versions. It's probably time to drop that support now.

bbatsov:
I guess I missed how this happened. I always had in mind only support for Emacs 24+.

PR #239, opened 2016-06-24
Make fci-mode line visible in light theme

wbolster:
Use the proper color definition for the very thin line shown by `fci-mode`. The `s-line` definition is intended for exactly this purpose. The result is a visible line in the light theme, which is currently invisible.

### DIFF
--- a/solarized.el
+++ b/solarized.el
@@ -2138,7 +2138,7 @@ customize the resulting theme."
      `(cua-global-mark-cursor-color ,cyan)
      `(cua-overwrite-cursor-color ,yellow)
 ;;;;; fill-column-indicator
-     `(fci-rule-color ,base02)
+     `(fci-rule-color ,s-line)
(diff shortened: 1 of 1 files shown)


thomasf:
it's visible in the light theme already for me... 

I'm leaning towards recommending that you change the `fci-rule-width` customization instead of applying this change.

### what's changed

this pull request changes this:

image

into this:

image

As seen in the screenshot above the fci line connects to the header line which is `base02` which is a reason to keep it as is.


### fci-rule-width

I found that I have this in y init.el which makes the current coloring work better:

```
(setq fci-rule-width 2)
```

With the default 1 pixel fci line width and `base02`:

image

With the default 1 pixel fci line width and `s-line`:

image

### window separator lines

Personally (don't remember if this is themable or not) I use a thin `s-line` colored as a window separator which makes this view very confusing

image

PR #243, opened 2016-08-24
magit: use colors in popup buffers

tarsius:
This is the theme I personally use, which is why I made the necessary changes myself when I released Magit v2.1.

Shortly after you merged that, you also changed the faces used in Magit's popup buffers to be only black&white instead of using colors as I had intended. I pointed that out to you but you never responded. See b0b8cb1d5c2b1836fbc764ee4279cccec0510aac.

I have reverted that commit for my own use. I also use that patched version when creating Magit screenshots. I intend to add more screenshots, but first I am making another attempt to convince you to revert your change from a year ago.

Black&white version:

bland

After applying this commit:

color

I hope you will apply this commit as is, but if this is too fruit-salady for you, then I guess I can compromise. If necessary I will present my arguments for each of these changes (reverts), but I am hoping I can avoid that work.

So what do you think?

### DIFF
--- a/solarized.el
+++ b/solarized.el
@@ -1322,11 +1322,11 @@ customize the resulting theme."
      `(magit-diffstat-added              ((t (:foreground ,green))))
      `(magit-diffstat-removed            ((t (:foreground ,red))))
 ;;;;;; popup
-     `(magit-popup-heading             ((t (:foreground ,base1 :weight normal))))
-     `(magit-popup-key                 ((t (:foreground ,base1 :weight bold))))
-     `(magit-popup-argument            ((t (:foreground ,base1 :weight bold))))
-     `(magit-popup-disabled-argument   ((t (:foreground ,base01 :weight normal))))
-     `(magit-popup-option-value        ((t (:foreground ,base1 :weight bold))))
+     `(magit-popup-heading             ((t (:foreground ,yellow  :weight bold))))
+     `(magit-popup-key                 ((t (:foreground ,green-d :weight bold))))
+     `(magit-popup-argument            ((t (:foreground ,cyan-d  :weight bold))))
+     `(magit-popup-disabled-argument   ((t (:foreground ,base01  :weight normal))))
+     `(magit-popup-option-value        ((t (:foreground ,cyan    :weight bold))))
 ;;;;;; process
      `(magit-process-ok    ((t (:foreground ,green :weight bold))))
      `(magit-process-ng    ((t (:foreground ,red   :weight bold))))


bbatsov:
I'm fine with the changes.

tarsius:
Thanks!

thomasf:
:+1:

mrkkrp:
tarsius-colors-suck

While the new colors may looks better for light version of the theme, they do look much worse for dark version. In fact, it's hard to read Magit popup now, as you can probably see on this screenshot. What we should do is to change the colors for dark version so text is contrast enough to be easily read.

thomasf:
yeh,.we should remove the dark colors, they arent particualy compatible with both themes when not  used as fg/bg combo..

thomasf:
It seems like I will semi revert this change again. I will post screenshot here before pushing anything.

thomasf:
this is with some colors but less than after this merge:
image

```el
     `(magit-popup-heading             ((t (:foreground ,yellow  :weight bold))))
     `(magit-popup-key                 ((t (:foreground ,base1 :weight bold))))
     `(magit-popup-argument            ((t (:foreground ,green  :weight bold))))
     `(magit-popup-disabled-argument   ((t (:foreground ,base01  :weight normal))))
     `(magit-popup-option-value        ((t (:foreground ,cyan    :weight bold))))

```

thomasf:
blue is possible a better active color than cyan is (in previous comment).

image

image

tarsius:
So you want to change `magit-popup-key` and `magit-popup-argument`. Great that's the ones I was willing to compromise on anyway ;-)

In the light theme I still like green-d for the keys better. It's almost black but with a hint of green. But all-black works for me too.

For arguments I originally picked cyan-d because it is almost like cyan but not quite. Cyan is used for the option-value because it is a string and those are cyan across solarized. I wanted the argument to be slightly different, because these two thinks are actually a unit, the argument and the value are of course passed to git as one argument.

Using green or blue for the value breaks that similarity with the argument. I find that a bit to fruitsalady, but more importantly these two colors are used for remote and local branches across magit. Now that the branch popup displays the names of branches in section headers and as the value of some variables, I think it would make sense not to use these colors for anything else.

Just using bold default for the active arguments would work for me. But I would like the arguments to not look exactly as the keys, so maybe those could be made "black with a hint of green" again? At least for the light theme, for the dark theme that doesn't work that well indeed.

tarsius:
What about only changing the colors for the dark theme because only here the contrast wouldn't be big enough?

```lisp
     `(magit-popup-heading             ((t (:foreground ,yellow  :weight bold))))
     `(magit-popup-key                 ((((class color) (background light))
                                         (:foreground ,green-d :weight bold))
                                        (((class color) (background dark))
                                         (:foreground ,base1 :weight bold))))
     `(magit-popup-argument            ((((class color) (background light))
                                         (:foreground ,cyan-d :weight bold))
                                        (((class color) (background dark))
                                         (:foreground ,cyan :weight bold))))
     `(magit-popup-disabled-argument   ((t (:foreground ,base01  :weight normal))))
     `(magit-popup-option-value        ((t (:foreground ,cyan    :weight bold))))
```

thomasf:
Generally I don't like having the `.*-d` colors used on normal backgrounds
since the contrast ratios aren't really tuned for that.

**ramble**

In a perfect world I would have the time to define more usable extra accent
colors which matches better to the normal background for light/dark, I've made
a few attempts but it's a very time consuming process checking against
different displays which is needed because of the Solarized palette already has
a thin contrast space available to do something with and for some not so good
displays it's sometimes impossible to even see the difference when the contrast
difference isn't large enough.

tarsius:
> Generally I don't like having the .*-d colors used on normal backgrounds since the contrast ratios aren't really tuned for that.

Then I would suggest using this:

```lisp
     `(magit-popup-heading             ((t (:foreground ,yellow  :weight bold))))
     `(magit-popup-key                 ((t (:foreground ,base1   :weight bold))))
     `(magit-popup-argument            ((t (:foreground ,cyan    :weight bold))))
     `(magit-popup-disabled-argument   ((t (:foreground ,base01  :weight normal))))
     `(magit-popup-option-value        ((t (:foreground ,cyan    :weight bold))))
```

Which is basically what you suggested above, except that arguments and their values now look exactly the same. You suggested using green or blue, which I think we should avoid because those colors are used for branches as explained above.

thomasf:
I had a feeling that the branch color code-thing was a thing you wanted to have.. I will probably commit your above version soon after I've tried it for a few hours. I think it's safe to assume that the user doesn't really need to heavily informed that the argument and it's value are different entities.

tarsius:
> I think it's safe to assume that the user doesn't really need to heavily informed that the argument and it's value are different entities.

I agree, that's why I originally used very similar colors, but not quite. Aside from the issue that the colors you suggest should be reserved for branches, using blue or green would additionally emphasis *too much* that these are different entities I think.

Soooooo in summary, if the dark variant I originally used for the active arguments is causing issues, then just using the same color as for the argument values is perfectly fine with me.

thomasf:
Since I'm drinking wine and githubbing right now anyways I'll just th

mrkkrp:
@thomasf Looks much better and more consistent with the other colors, thanks!

PR #244, opened 2016-09-14
use only background color for region selection

lboogaard:
I've found it annoying at times that selecting a region overwrites all the syntax highlighting (by setting a foreground color).

Would it be an option to set only a background color for a selected region? I've experimented a bit and I think `base02` would be the natural choice, `base01` is a bit too dark. Note that `base02` is also used for `secondary-selection`; not sure if that is a problem for people?

### DIFF
--- a/solarized.el
+++ b/solarized.el
@@ -423,7 +423,7 @@ customize the resulting theme."
                                 :box (:line-width 2 :color ,s-header-line-bg
                                                   :style unspecified)
                                 ))))
-     `(region ((,class (:foreground ,base03 :background ,base1))))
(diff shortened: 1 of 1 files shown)


lboogaard:
With some screens: 

screen shot 2016-09-14 at 15 39 21
screen shot 2016-09-14 at 15 39 48

thomasf:
It will conflict with a lot of other things as well.

lboogaard:
I guess `base01` won't give any conflicts (or `base1` for that matter). But I'm not convinced that is actually better than losing the syntax highlighting to improve contrast.

PR #246, opened 2016-11-07
Use base02 for org-clock-overlay face

marsam:
Use base02 for `org-clock-overlay` face used by `org-clock-display`

**Original**

screenshot_20161107_120002

**New**
screenshot_20161107_120134


thomasf:
Seems to be good..

PR #256, opened 2017-02-28
Corrected README to set scale rather than absolute height

### LINKED ISSUE #251: Problem disabling heading size changes
Setting the variables solarized-height-plus to 1 as instructed in the readme makes the height of all LaTeX headings exactly 1, it does not avoid scaling. This makes the headlines invisible.

fwkroon:
Setting the height attribute to an integer sets the font height to an absolute value in units of 1/10 point. This patch fixes #251 by changing the README to use floats, thus setting the font scale instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,11 +60,11 @@ elisp code:
 (setq solarized-scale-org-headlines nil)
 
 ;; Avoid all font-size changes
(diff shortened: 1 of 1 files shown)


thomasf:
thanks

PR #258, opened 2017-03-29
Make org-block-begin-line inherit from org-meta-line

alphapapa:
In org-mode itself, `org-block-begin-line` inherits from `org-meta-line`, as does `org-block-end-line`, which ensures that the begin and end lines look the same.  Somehow that has not been the case in solarized-theme; this restores it.  

Thanks for your work on this theme!  Been using it for years and there's nothing better.  :)

### DIFF
--- a/solarized.el
+++ b/solarized.el
@@ -1610,7 +1610,7 @@ customize the resulting theme."
      `(org-agenda-done ((,class (:foreground ,base01 :slant italic))))
      `(org-archived ((,class (:foreground ,base01 :weight normal))))
      `(org-block ((,class (:foreground ,base01))))
-     `(org-block-begin-line ((,class (:foreground ,base01 :slant italic))))
+     `(org-block-begin-line ((,class (:inherit 'org-meta-line :foreground ,base01 :slant italic))))
      `(org-checkbox ((,class (:background ,base03 :foreground ,base0
                                           :box (:line-width 1 :style released-button)))))
      `(org-code ((,class (:foreground ,base01))))


thomasf:
where is this change visible?

alphapapa:
For me it is visible in `#+BEGIN_SRC` lines vs. `#+END_SRC` lines: the begin line is not bold because it does not inherit from `org-meta-line`.  They are supposed to look the same.

thomasf:
hmm. they should not be bold at all if i'm not missing something..

both my start/end example/src's are just italic without any bold like this.. 

image

thomasf:
Maybe we use different org mode versions or something.. It's probably for the best to just add org-meta-line and org-end-block to the theme as well.. The theme doesn't use face inheritance because there were too many situations where changes cascaded into the wrong places..

alphapapa:
Yeah, I'm still using an older version.  I need to upgrade it sometime...  :)

I looked up the face definitions in `org-faces.el`, and `org-block-begin-line` and `org-block-end-line` are both set to inherit from `org-meta-line`.  So assuming that you want both the begin and end lines to look the same in this theme, I guess they should just inherit from `org-meta-line` (i.e. I guess they shouldn't be mentioned in the theme, since they inherit by default), and `org-meta-line` should be the one that gets themed.  Does that make sense?

thomasf:
in my org mode. end-line inherits from begin-line, seems more correct..

Does adding this work out for you?
```el
     `(org-block-end-line ((,class (:inherit org-block-begin-line))))
```

alphapapa:
I agree, that does make even more sense.  Sorry that my using an older Org is complicating this a bit.  You should definitely do whatever is right for the current Org version, and I can work around any issues on my end while I'm still using the older one.

In fact, I wonder if this is a complete non-issue on current Org, and maybe this whole issue is because I'm using the older version.  D:

PR #259, opened 2017-04-14
Set indent of solarized-with-color-variables to defun

alphapapa:
This sets the indent declaration to `defun` so that it's indented like other macros.  Thanks.


alphapapa:
I don't know why I did this twice.  Must have gotten confused after reloading Emacs...

PR #260, opened 2017-04-14
Set org-meta-line instead of org-block-begin-line

alphapapa:
Previously discussed at https://github.com/bbatsov/solarized-emacs/pull/258

I've upgraded Org to 9.0.5 and confirmed that this fix is needed.  Thanks.


thomasf:
I guess this should work

PR #261, opened 2017-04-14
solarized.el: Set indent of solarized-with-color-variables

alphapapa:
This adjusts the indentation of the macro so that uses of it are indented like normal macros.  

It might be a good idea to reindent the source files after this in another commit.  Let me know if you'd like me to do this.

Thanks.


alphapapa:
Ping?  :)

thomasf:
Right, it seems that indentation is slightly wrong here and there..

PR #271, opened 2017-08-25
Add background class variables

blaenk:
These are like the exposed `class` variable which we can use in specs as `,class`, except they also contain a display element specifying the background variant to apply to.

Note that for 6144e22 I make use of them, but I'm not sure if it should be done, since notice that those specs don't include the `(min-colors 89)` spec. If that was unintentional, then using these class variables is correct, but not if it was intentional to omit it.

### DIFF
--- a/solarized.el
+++ b/solarized.el
@@ -332,9 +332,10 @@ customize the resulting theme."
      `(dired-symlink ((,class (:foreground ,cyan :weight normal :slant italic))))
      `(dired-warning ((,class (:foreground ,orange :underline t))))
 ;;;;; dired-async
-     `(dired-async-message ((,class (:background ,(if (eq variant 'light) yellow-l yellow) ))))
-     `(dired-async-mode-message
-       ((,class (:background ,(if (eq variant 'light) red-l red) ))))
+     `(dired-async-message ((,light-class (:background ,yellow-l ))
+                            (,dark-class (:background ,yellow ))))
+     `(dired-async-mode-message ((,light-class (:background ,red-l))
+                                 (,dark-class (:background red))))
 ;;;;; dired-efap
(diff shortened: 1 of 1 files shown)


blaenk:
This allows the following kind of refactorings:

``` diff
diff -u --label /home/blaenk/code/emacs/solarized-emacs/solarized.el --label \#\<buffer\ solarized.el\> /home/blaenk/code/emacs/solarized-emacs/solarized.el /tmp/buffer-content-cz2fQu
--- /home/blaenk/code/emacs/solarized-emacs/solarized.el
+++ #<buffer solarized.el>
@@ -332,7 +332,8 @@
      `(dired-symlink ((,class (:foreground ,cyan :weight normal :slant italic))))
      `(dired-warning ((,class (:foreground ,orange :underline t))))
 ;;;;; dired-async
-     `(dired-async-message ((,class (:background ,(if (eq variant 'light) yellow-l yellow) ))))
+     `(dired-async-message ((,light-class (:background ,yellow-l ))
+                            (,dark-class (:background ,yellow ))))
      `(dired-async-mode-message
        ((,class (:background ,(if (eq variant 'light) red-l red) ))))
 ;;;;; dired-efap

Diff finished.  Fri Aug 25 13:14:03 2017
```

blaenk:
Just pushed 92c26e3 which eliminates explicit/manual conditions on the `variant` variable.

I don't have any of the packages that the faces are for, but I was able to confirm that the specs are being set by checking the symbols' property lists.

blaenk:
Heads up: I just force pushed to fix a typo.

blaenk:
I made a final force push to use `append` instead of `-concat`, since `append` is more to the point, implemented in C, and already used in the file.

PR #276, opened 2017-11-14
add auto-dim-other-buffers-face

kenda:
This PR adds the face for `auto-dim-other-buffers-mode` which dims the currently inactive buffers by using a different background colour.


thomasf:
thanks

PR #279, opened 2017-12-02
Add faces for spaceline

andersjohansson:
The highlight face with yellow background may be considerd ugly but I think it looks ok.
screenshot of modeline
(screenshot using spaceline-all-the-icons)


bbatsov:
👍

PR #280, opened 2017-12-02
Add faces for all-the-icons.

andersjohansson:
Add faces for the three related packages
- [all-the-icons](https://github.com/domtronn/all-the-icons.el)
- [all-the-icons-dired](https://github.com/jtbm37/all-the-icons-dired)
- [spaceline-all-the-icons](https://github.com/domtronn/spaceline-all-the-icons.el)

This uses the extra light and dark color faces. I think icons are a valid case for using some "special" colours. I'm not entirely sure about the use of the the high-contrast, low-contrast faces. Perhaps it should just be kept consistent so that for instance a file type like pdf always gets a dark red icon, both when using solarized-light and solarized-dark.
The all-the-icons faces are defined with different colors for light and dark backgrounds though, but the usage there is a bit inconsistent: https://github.com/domtronn/all-the-icons.el/blob/master/all-the-icons-faces.el


bbatsov:
Can we see some screenshot with the proposed faces?

andersjohansson:
Of course!
First the mode-line stuff. 
weather
The temperature icon gets a calculated color (not a themable face) so it can be hard to see, but it’s not the problem of solarized. (This screenshot also incidentally shows the hours of our, luckily sunny, 6-hour day at the winter solstice here in Uppsala, Sweden 🙂)

Then all the icons for files and modes. I just iterated through the lists so these are the icons with colors which is normally output:
files-light
files-dark
modes-light
modes-dark

To me it generally looks better with the solarized-light.I don’t know if it perhaps is a bad idea to use the `-lc` and `-hc`-faces, or if some individual colors should be tweaked (just using `-d` or `-l` colors instead of `-hc`, `-lc`) . As I said, the definitions in `all-the-icons` are not really consistent.

Finally, a small example of how it looks in dired:
icons-dark-light

Issue #134, opened 2014-11-07
Ediff faces work poorly with dark theme


benkuhn:
Using ediff with the `solarized-dark` theme, the backgrounds of diff regions can get quite bright. Combined with dark text (e.g. comments) superimposed, this can make some text very hard to read:

![screen shot 2014-11-07 at 2 03 16 pm](https://cloud.githubusercontent.com/assets/304831/4961027/f5bd6dc8-66c9-11e4-8659-d0c838f4cda5.png)

Here it is with the light theme for comparison:
![screen shot 2014-11-07 at 2 06 12 pm](https://cloud.githubusercontent.com/assets/304831/4961058/595ff9b8-66ca-11e4-8728-a04fcc55e345.png)

thomasf:
I think that these colors need to be calculated at theme creation time to allow the various levels of highlight to work properly.

wyuenho:
This can be closed now right?

terzievk:
I really like the solarized theme. However I have a problem with ediff too - mainly with how translucent the marking is, which results in unreadable comments. Here is a screenshot:

![screenshot from 2018-09-26 00-18-56](https://user-images.githubusercontent.com/17403291/46044155-a5d44180-c122-11e8-8722-88dc7b040174.png)
So my question is: Is there a way to set a default go to theme to be set for such problematic occasions?

P.S: I don't know if this is mentioned elsewhere but after loading another theme and then loading the solarized theme back it sort of fixed itself *weird*. I don't think this sounds clear enough so here are some screenshots:
- here the solarized theme is set by default:

![screenshot from 2018-09-26 00-18-56](https://user-images.githubusercontent.com/17403291/46044369-40348500-c123-11e8-8d1f-e233796d1614.png)

- here I load the deeper-blue theme:

![screenshot from 2018-09-26 00-19-12](https://user-images.githubusercontent.com/17403291/46044461-7a9e2200-c123-11e8-8c12-209842ca81c5.png)

- now the marking is opaque and easily readable:

![screenshot from 2018-09-26 00-19-15](https://user-images.githubusercontent.com/17403291/46044483-8d185b80-c123-11e8-863b-515050511d72.png)

- now I load the solarized-dark theme:

![screenshot from 2018-09-26 00-19-28](https://user-images.githubusercontent.com/17403291/46044509-9acde100-c123-11e8-95c4-da8cec525ca2.png)

- the result is sort of mix between the two:

[...]

PR #282, opened 2017-12-25
Fix #194 Unified diff diff-hl ediff and smerge faces

### LINKED ISSUE #192: smerge-mode support
Support to the built-in Emacs minor mode "smerge" would be great. Thanks!

### LINKED ISSUE #194: Various diff/merge modes are broken
Something entirely new needs to be done for this to work.. 
This is a meta issue.

Related issues: 
- #192 
- #191 
- #134

wyuenho:
This should fix #192 and #134. In this PR I've unified the colors as much as I could. Some modes have extra faces for changed blocks, in contexts where there are only 2 major color values (red and green), I've used yellow to indicate changes. When there's more, I've used blue-lc background and cyan-hc foreground to make the text more legible.

When outside of changed blocks (added and deleted), the foreground is highlighted with a slightly higher contrast color (base3), for legibility again. Font locking weights and slants are removed in all hunks to improved legibility.

For diff and ediff, refined blocks are highlighted by using the high contrast version of the background color on the foreground. They are underlined too.

For ediff and smerge, anything other than the hunks are only subtlely highlighted to minimize distraction.

Oh, did I mention there's smerge theme now? Merry Christmas!

screen shot 2017-12-25 at 20 02 28
screen shot 2017-12-25 at 20 03 02
screen shot 2017-12-25 at 20 05 36
screen shot 2017-12-25 at 20 06 18
screen shot 2017-12-25 at 21 26 53
screen shot 2017-12-25 at 21 28 31

### DIFF
--- a/solarized.el
+++ b/solarized.el
@@ -613,118 +613,45 @@ customize the resulting theme."
      `(custom-button-unraised ((,class (:inherit underline))))
      `(custom-button-pressed-unraised ((,class (:inherit custom-button-unraised :foreground ,magenta))))
 ;;;;; diff
-     `(diff-added   ((,class (:foreground ,green))))
-     `(diff-changed ((,class (:foreground ,blue))))
-     `(diff-removed ((,class (:foreground ,red))))
-     `(diff-refine-added
-       ((,light-class
-         (:background ,(solarized-color-blend "#ddffdd" green 0.7)))
-        (,dark-class
-         (:background ,(solarized-color-blend "#446644" green 0.7)))))
-     `(diff-refine-changed
-       ((,light-class
-         (:background ,(solarized-color-blend "#ddddff" blue 0.7)))
-        (,dark-class
-         (:background ,(solarized-color-blend "#444466" blue 0.7)))))
-     `(diff-refine-removed
-       ((,light-class
-         (:background ,(solarized-color-blend "#ffdddd" red 0.7)))
-        (,dark-class
-         (:background ,(solarized-color-blend "#664444" red 0.7)))))
+     `(diff-added ((,class (:background ,green-lc :foreground ,base3 :weight normal :slant normal))))
+     `(diff-changed ((,class (:background ,yellow-lc :foreground ,base3 :weight normal :slant normal))))
+     `(diff-removed ((,class (:background ,orange-lc :foreground ,base3 :weight normal :slant normal))))
+     `(diff-refine-added ((,class (:foreground ,green-hc :underline t))))
+     `(diff-refine-changed ((,class (:foreground ,yellow-hc :underline t))))
+     `(diff-refine-removed ((,class (:foreground ,orange-hc :underline t))))
      `(diff-header  ((,class (:background ,base03))))
      `(diff-file-header
        ((,class (:background ,base03 :foreground ,base0 :weight bold))))
 ;;;;; diff-hl
-     `(diff-hl-change ((,class (:background ,blue-lc  :foreground ,blue-hc))))
-     `(diff-hl-delete ((,class (:background ,red-lc  :foreground ,red-hc))))
-     `(diff-hl-insert ((,class (:background ,green-lc  :foreground ,green-hc))))
-     `(diff-hl-unknown ((,class (:background ,cyan-lc   :foreground ,cyan-hc))))
+     `(diff-hl-change ((,class (:background ,blue-lc :foreground ,blue-hc))))
+     `(diff-hl-delete ((,class (:background ,orange-lc :foreground ,orange-hc))))
+     `(diff-hl-insert ((,class (:background ,green-lc :foreground ,green-hc))))
+     `(diff-hl-unknown ((,class (:background ,yellow-lc :foreground ,yellow-hc))))
 ;;;;; ediff
-     `(ediff-fine-diff-A ((,class (:background ,orange-lc))))
-     `(ediff-fine-diff-B ((,class (:background ,green-lc))))
-     `(ediff-fine-diff-C ((,class (:background ,yellow-lc))))
-
-     `(ediff-current-diff-C ((,class (:background ,blue-lc))))
-
-     `(ediff-even-diff-A ((,class (:background ,base01
-                                               :foreground ,base3 ))))
-     `(ediff-odd-diff-A ((,class (:background ,base01
-                                              :foreground ,base03 ))))
-     `(ediff-even-diff-B ((,class (:background ,base01
-                                               :foreground ,base03 ))))
-     `(ediff-odd-diff-B ((,class (:background ,base01
-                                              :foreground ,base03 ))))
-     `(ediff-even-diff-C ((,class (:background ,base01
-                                               :foreground ,base0 ))))
-     `(ediff-odd-diff-C ((,class (:background ,base01
-                                              :foreground ,base03 ))))
-
-;;;;;; alternative ediff (not finished)
-     ;; `(ediff-fine-diff-A ((,class (
-     ;;                               :background ,(solarized-color-blend blue base03 0.25))
-     ;;                              )))
-     ;; `(ediff-fine-diff-B ((,class (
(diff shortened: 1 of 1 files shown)


thomasf:
I will try to take a look at this soon. Thanks for spending the effort.

wyuenho:
How does this look?

wyuenho:
How does this PR look now?

bbatsov:
Thanks for working on this! I really appreciate it! 🙇And sorry for the slow response from me!

tarsius:
This makes diffs in Magit look much much worse. When I saw the effect of this in a Magit diff it didn't even occur to me that the faces might have been changed deliberately, I assumed that I somehow ended up activated some mode that is not compatible that with Magit and started bisecting my init file.

That's how much I dislike this new look. Obviously some people prefer this new look, but this is not a change that should be made with so little discussion.

I spend a lot of time tweaking the old colors. Could we please not throw that all out of the window?

So could we please revert this and then talk about it?

> In this PR I've unified the colors as much as I could.

Additionally you have changed "fine diffs" from using fine-tuned background colors to underlines. I think that looks much worse. That's a personal preference I supposed, but since that what we did for a few years I suspect a few other users have grown fund of it.

tarsius:
My thoughts on how I think diff faces (and some others) are best themes can be found [here](https://magit.vc/manual/magit/Theming-Faces.html). It ends with:

> Nobody said this would be easy. If your theme restricts itself to a certain set of colors, then you should make an exception here. Otherwise it would be impossible to make the diffs look good in each and every variation. Actually you might want to just stick to the default definitions for these faces. You have been warned. **Also please note that if you do not get this right, this will in some cases look to users like bugs in Magit - so please do it right or not at all.**

Well I didn't expected to be one of those users, but here we are.

I am using Solarized light myself and it was me who has themed its faces used in Magit diffs. I did not do so for other diff related packages. I rarely use Edif and `diff-mode` even less.

I am all for making diffs look uniform across packages but did not do it for two reasons:

- I rarely use Edif and `diff-mode` even less. Smerge already looks fine I think.
- I did not want to change how they look for people who are already used to them looking differently.

I did however feel it was justified in deciding how diffs should look in Magit. So by all means make diffs look uniform across packages, but I would suggest you do so by porting the Magit theming to other modes.

wyuenho:
>> This makes diffs in Magit look much much worse.

I haven't touched any magit faces, are your magit faces derived from what's defined here? I've left all my magit faces to their default and I happen to think magit is the one package where no face intervention is needed. They spent a lot of time tuning the color to suit every situation.

>> I spend a lot of time tweaking the old colors. Could we please not throw that all out of the window?

As you can see, the old color were creating a lot of problems due to multiple font locks being in place and lack of contrast. If you have any specific complaint about this new change, can you provide more context, preferably with some before and after screenshots please?

I've laid out my reasons for changing the faces in the description, so if you will, can you please respond to it please?

thomasf:
Yeah, sorry I haven't had time to look at this issue.. diff/merge modes are indeed a complicated matter.. I agree that it maybe looks a bit too much. A reason for setting as few properties as possible (normally only background) for merge modes is to not hide normal buffer fontification. 
I think we actually had a more aggressive style earlier, in some ways resembling what done here but it was toned down in favor of compatibility with other colors..

thomasf:
Personally I feel that smerge is totally broken now,, it looks like a mess. I prefer to see the code presented as close to how it looks in normal editing as possible.

The various formattings and colors are kind of cryptic looking as well:

image
image

tarsius:
> I haven't touched any magit faces, are your magit faces derived from what's defined here?

Most faces are not derived, but hunks are refined (in `magit-diff-update-hunk-refinement`) using `diff-refine-hunk`, so the `diff-refine-*` faces are used in Magit diffs.

> I've left all my magit faces to their default and I happen to think magit is the one package where no face intervention is needed.

Have enabled hunk refinement in Magit buffers using `(setq magit-diff-refine-hunk t)`? Do you like that look or is that too little contrast?

> They spent a lot of time tuning the color to suit every situation.

They is me.

> As you can see, the old color ...

No I can't - you only posted the after screenshots.

> ... were creating a lot of problems due to multiple font locks being in place ...

I don't know what "multiple font locks being in place" means.

> ... and lack of contrast.

I acknowledge that this is a legitimate concern. But Solarized is a low-contrast theme, so this should not be a primary concern in this case. An option "high-contrast diffs" could be added, but it should not be the default.

> If you have any specific complaint about this new change, can you provide more context, preferably with some before and after screenshots please?

Frankly it is your job to do that because you are the one you proposed this change. But I will do that at a later time. I was very excited to get a lot of work done today and this is a major distraction.

> I've laid out my reasons for changing the faces in the description, so if you will, can you please respond to it please?

My current understanding is that you want consistency and contrast. I want "semantic colors" instead (red=remove, green=added, grayish=context, yellow=common-ancestor) and want to use a different shades of the same colors for the fine changes.

You have not explained why it is an improvement to use the same color (blue) for fine changes in both added and removed lines in the case of merge conflicts (smerge), but to use underline in diffs instead. And how that improves consistency (previously different shades were used in both cases).

tarsius:
@thomasf I agree with your assessment.

thomasf:
I have also spent many hours on diff mode(s) faces :)  

The simplicity of of the Solarized palette in general and this Emacs theme to [an even larger extent](https://github.com/bbatsov/solarized-emacs/blob/master/DEV-GUIDE.md) than many other editor themes means that we often sacrifice some color coded information and focus on the most important aspects... In some cases it has taken  several years of repeated simplifications to get to how certain things looks..

thomasf:
I do agree that s-merge probably should have faces defined, I use it all the time but that can be done separatley.  I'm reverting this change for now.

wyuenho:
> I don't know what "multiple font locks being in place" means.

I was referring to #134 . That's a problem with having faces applied on top of faces. It's very confusing, so I opted to use the "more aggressive approach" as @thomasf described, namely override all the weight/slant/underline BS that may be there due to linter or major mode syntax highlighting. A diff is a diff, no syntax highlighting is needed unless it consists of only 2 colors, which I can't really tell without a shit ton of work and massively slow down font locking.

> Have you enabled hunk refinement

No. I'll check that.

> The simplicity of of the Solarized palette in general

I'm beginning to think the solarized palette is appropriate for dealing with diffs, which requires probably a dozen monochrome colors separated into 3 main groups. I'd be happy to adapt magit's colors with some slight tweak to ediff/diff/diff-hl/smerge for solarized.

> Frankly it is your job to do that because you are the one you proposed this change.

Mind you this PR has been sitting here for almost 3 months and I personally have not much issue with it. At some point you are just going to have to call 

@thomasf The color of your screen shots look totally different from mine for the added and removed colors. I have also not checked the common ancestor color as I've actually never seen it in real life. I'll make a quick PR to take out the smerge colors for now if no one objects, but I'd like the rest to stay, as they are mostly inspired by most other editors' color schemes, and github's.

I'll take a look at magit hunk.

tarsius:
> I have also spent many hours on diff mode(s) faces :)

I was talking about the Magit faces when I said "they is me". Didn't want to imply I did all the work.

thomasf:
I agree that the the ediff faces are quite bad in general.

You should always verify changes with as many optional settings as possible for everything. 
And btw, I have `magit-diff-refine-hunk` set to `'all`. 

The review scope for this kind of change is large. I would prefer one pull request per mode as a starting point.
You could maybe start with a suggestion for `ediff` or `smerge` with before/after screenshots?

thomasf:
> I'm beginning to think the solarized palette is inappropriate for dealing with diffs, which requires probably a dozen monochrome colors separated into 3 main groups. I'd be happy to adapt magit's colors with some slight tweak to ediff/diff/diff-hl/smerge for solarized.

Yes, this is a big issue for colored diffs, there basically no any room in the full contrast difference between darkest and brightest to fit new intermediate colors without it quickly becoming very hard to read.

Issue #18, opened 2012-04-14
Incorrect Colours in Terminal


zmanji:
I'm not sure if this is a bug of solarized-emacs or my fault but when I load up emacs in a terminal window this is what i get:

![](http://f.cl.ly/items/3l0w0e1V1y1y1H3U0A2z/Screen%20Shot%202012-04-13%20at%2011.59.59%20PM.PNG)

However, my terminal has 256 color support and `M-x list-colors-display` displays all of the colours correctly. I don't know how to diagnose the issue further to determine where the problem lies.

andersjohansson:
@sellout has been doing some work on merging the unified-branch of the other theme (https://github.com/sellout/emacs-color-theme-solarized/) and creating tests recently (in the last 24 hours). I don't know how that relates to your work @thomasf, but I guess it could be useful to unite these efforts. I really like the "unified"-approach, and have been using this theme mostly because it supports more faces. Having the two themes merged in some way would also be quite cool.

mikefaille:
I personnaly have this problem with kmscons and tty terminal (only) using Linux.
Note : I think tty can be usefull to recover systems. (no more use case for me)

thomasf:
@andersjohansson  The version of solarized is a bit more complex because it's really an extended solarized in it's current high color only theme.

My previous attempt at doing this was unsuccessful because I was not satisfied with the preciseness of the code when doing it with the basic emacs lisp code.

Next time I will use pattern matching to build the facedefs. The target is to only depend on dash.el. I was actually holding off this work while waiting for dash.el to get support for plists which it now has!

mikefaille:
@thomasf I personnaly pass a lot of time to find.. Why Wombat OEM  work always with load theme  ? ! 
Emacs OEM theme set is only theme who always work using : kmscons --palette solarized
or just kmscons (without palette)
It's cool because... Wombat is really simple : http://git.savannah.gnu.org/cgit/emacs.git/tree/etc/themes/wombat-theme.el?h=emacs-24

Why emacs theme can be complicated ? It's because color interpolation between desired and current env. palette limitation.
If I understand by experience, emacs seems to interpolate colors very well by default. 
I think this is documented too (english isn't my 1st language and I try to don't use Google translate to pratice..) : https://www.gnu.org/software/emacs/manual/html_node/efaq/Colors-on-a-TTY.html

My suggestion : 
Make simple solarized theme using no color interpolation.

thomasf:
[...]

PR #283, opened 2017-12-27
Add option to use term colors.

beauby:
Ref: #18.

### DIFF
--- a/solarized.el
+++ b/solarized.el
@@ -123,6 +123,11 @@ Related discussion: https://github.com/bbatsov/solarized-emacs/issues/158"
   :type 'boolean
   :group 'solarized)
 
+(defcustom solarized-termcolors nil
+  "Use terminal color scheme instead of the 256 colors palette."
+  :type 'boolean
+  :group 'solarized)
+
 ;;; Utilities
 
 ;;;###autoload
@@ -133,11 +138,13 @@ COLOR1 and COLOR2 should be color names (e.g. \"white\") or RGB
 triplet strings (e.g. \"#ff12ec\").
 
 Alpha should be a float between 0 and 1."
-  (apply 'color-rgb-to-hex
-         (-zip-with '(lambda (it other)
-                       (+ (* alpha it) (* other (- 1 alpha))))
-                    (color-name-to-rgb color1)
-                    (color-name-to-rgb color2))))
+  (if solarized-termcolors
+      (if (< alpha 0.5) color2 color1)
+      (apply 'color-rgb-to-hex
+             (-zip-with '(lambda (it other)
+                           (+ (* alpha it) (* other (- 1 alpha))))
+                        (color-name-to-rgb color1)
+                        (color-name-to-rgb color2)))))
 
 ;;; Setup Start
 (defmacro solarized-with-color-variables (variant &rest body)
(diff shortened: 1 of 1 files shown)


bbatsov:
Can you show us how the degraded palette looks?

solarized.el
```diff
@@ -116,6 +116,11 @@ Related discussion: https://github.com/bbatsov/solarized-emacs/issues/158"
   :type 'boolean
   :group 'solarized)
 
+(defcustom solarized-termcolors nil
```
bbatsov: Shouldn't we detect the capabilities of the terminal?

beauby: AFAIK, the main issue is that a 256 color capable term emulator will be recognized as such, and those 256 colors do not contain the exact solarized palette, although the terminal might be configured with the official 16 colors palette as base colors. I may be wrong though.

thomasf: Yes, detection should not be trusted as the single source of truth.

bbatsov: I see.

solarized.el
```diff
@@ -126,11 +131,13 @@ COLOR1 and COLOR2 should be color names (e.g. \"white\") or RGB
...
 
 Alpha should be a float between 0 and 1."
-  (apply 'color-rgb-to-hex
-         (-zip-with '(lambda (it other)
-                       (+ (* alpha it) (* other (- 1 alpha))))
-                    (color-name-to-rgb color1)
-                    (color-name-to-rgb color2))))
+  (if solarized-termcolors
+      (if (< alpha 0.5) color2 color1)
+      (apply 'color-rgb-to-hex
```
bbatsov: This doesn't look right - the `else` should be outdented.

solarized.el
```diff
@@ -126,11 +131,13 @@ COLOR1 and COLOR2 should be color names (e.g. \"white\") or RGB
 triplet strings (e.g. \"#ff12ec\").
 
 Alpha should be a float between 0 and 1."
-  (apply 'color-rgb-to-hex
-         (-zip-with '(lambda (it other)
-                       (+ (* alpha it) (* other (- 1 alpha))))
-                    (color-name-to-rgb color1)
-                    (color-name-to-rgb color2))))
+  (if solarized-termcolors
+      (if (< alpha 0.5) color2 color1)
```
bbatsov: I'm not quite certain what this is supposed to do. Some comments might be useful.

0e39bf7b:
@bbatsov what should we to to merge this PR?

solarized.el
luisgerhorst: Is there any widespread terminal that correctly displays the solarized color codes? If not what about `(not (display-graphic-p)` as a heuristic? Either the user has set the solarized palette in their terminal or he did not. 
- If he has set the solarized palette it works fine if we use the heuristic. If we don't they look bad.
- If he has not set the solarized palette the colors look bad anyway, so it does not make a difference if we select the "fake" palette using the heuristic (assuming nobody wants to use the solarized theme in a 256 terminal where the colors are only displayed approximately).

thomasf:
Closing this because not compatible with current code.

This does not solve all use of color blending in the theme. Everything not using the standard 16 solarized colors needs to have individually tested for every change. 

I think it's just too much work given that there are 24bit compatible terminals that work in macos, linux and windows people should be nudged towards learning how to set that up instead.

PR #288, opened 2018-03-14
add highlight-parentheses (#241)

### LINKED ISSUE #241: highlight-parentheses support
what would make sense for highlighted parentheses colouring? it looks like this in the theme:

```
+;;;;; highlight-parentheses  (TODO)
+     `(hl-paren-colors '(,orange ,violet ,cyan ,green))
```

...but i'm not happy about how it looks. note that the defaults are even uglier. :)

LeandroCardoso:
Add support to highlight-parentheses package. Fix #241 . Using the same colors in the same order as rainbow-delimiters.


bbatsov:
Thanks!

PR #293, opened 2018-05-06
Adjust magit-blame faces to recent upstream changes

tarsius:
```
The hash/name/date/summary faces no longer inherit from
`magit-blame-heading'.  Instead they are combined with that or another
face like `:face (list BG-FACE %-FACE)' or in some cases they are not
combined with another face.

The heading face should only set the background and the
heading and the new `magit-blame-margin' should inherit from that.

It isn't necessary to theme that or the also new `magit-blame-dimmed'
here because these faces (and `magit-blame-heading') now `:weight
normal :normal' by default.
```

https://github.com/magit/magit/commit/1ee68c652f77ece1ce614e55ef082b9804660ed4
https://github.com/magit/magit/issues/2937

### DIFF
--- a/solarized.el
+++ b/solarized.el
@@ -1418,20 +1418,17 @@ customize the resulting theme."
      `(magit-bisect-skip ((t (:foreground ,yellow))))
      `(magit-bisect-bad  ((t (:foreground ,red))))
 ;;;;;; blame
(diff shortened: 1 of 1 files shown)


bbatsov:
👍

PR #294, opened 2018-05-26
Fix header and mode line box style value

wyuenho:
box style cannot be `unspecified`, but it can be `nil`. Specifying `unspecified` prevents configuration with `customize-face`.


thomasf:
IIRC unspecified have the benifit over nil that it doesnt override other faces which tries to set the property, I'm not sure if it's needed in the mode line though.

wyuenho:
Do you remember what they are? I only came across this because I was trying to turn off the box and the underline on the mode line.

PR #296, opened 2018-05-29
company mode changes

thomasf:
a work in progress.. Trying to align company mode with how my auto-complete-mode looks.

I have just switched to company mode only for go so I need to try it out with other modes as well to see if there are other facedefs specific to other modes which needs to be considered.

### (my) auto-complete-mode config with python
image

### current company-mode look

image

### current draft change

It's not really there yet but somwhat similar to ac-mode.

image


thomasf:
search/filter hl

image

thomasf:
if no-one complaings I will maybe try to modify auto-complete-mode as well.. There is a big difference though in that ac-mode providers have their own facedefs which I havent found while browsing company-mode ones so maybe it's not that easy to adapt ac-mode.

PR #303, opened 2018-06-18
Add: magit-filename face

alphapapa:
Chose violet because it doesn't seem to be already used in the Magit faces.

### DIFF
--- a/solarized.el
+++ b/solarized.el
@@ -1436,6 +1436,7 @@ customize the resulting theme."
      `(magit-blame-date      ((t (:foreground ,violet))))
 ;;;;;; references etc.
      `(magit-dimmed         ((t (:foreground ,base01))))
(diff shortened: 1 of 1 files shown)


thomasf:
It has the default foreground color now, right? I think that's enough, no need to spend more colors.

thomasf:
Since @tarsius has defined most of the magit face specs I don't think it's been overlooked(?)

tarsius:
I believe I added this face at the time in order to *remove* theming of filenames in certain contexts. The commit message doesn't why I made the change, but the default value, `((t :weight normal))`, indicates that all I wanted to do is for the filenames *not* to be displayed bold.

This doesn't seem to be necessary anymore (though I haven't looked very carefully), but I am going to keep the face anyway. Some people might already have themed it.

But this face should not be changed in popular themes to make it look special. That's not what it was intended for, quite the opposite.

PR #307, opened 2018-06-21
Add org-macro

grettke:
Adds org-macro, defining it slightly (one step) brighter than the normal text: noticeable, without interrupting the flow with a color.

### DIFF
--- a/solarized.el
+++ b/solarized.el
@@ -1724,6 +1724,7 @@ customize the resulting theme."
                                       :foreground ,blue))))
      `(org-link ((,class (:foreground ,yellow :underline t))))
      `(org-meta-line ((,class (:foreground ,base01 :slant italic))))
+     `(org-macro ((,class (:foreground ,s-base1))))
      `(org-sexp-date ((,class (:foreground ,violet))))
      `(org-scheduled ((,class (:foreground ,green))))
      `(org-scheduled-previously ((,class (:foreground ,cyan))))


thomasf:
I'm not sure where this is used, could you post a before/after screenshots?

grettke:
Sure is a before, after, and `describe-char` on it:

thomasf:
alright, lgtm

alphapapa:
This seems inconsistent with the idea of theming Emacs faces according to Solarized.  Org defines the macro face as being a distinct color, nearly yellow.  For Solarized, shouldn't it be themed as Solarized's yellow?  This is virtually de-fontifying that face, which isn't what users will expect.  If a user wants to make this face almost blend into regular text, shouldn't that be done in the user's personal config?

thomasf:
There is quite a lot of "de-fontifying" going on in this themes so it's not automatically seen as a bad thing. What is the problem of blending into regular text?  I dont remember but yellow is already used for links in org-mod (maybe)...

alphapapa:
> What is the problem of blending into regular text?

Well, it seems like the job of a color theme is to set the faces defined by packages according to a certain color scheme.  But if a theme takes a face which is defined by a package to stand out, and makes that face not stand out, that seems outside the purview of a color theme.  It also violates the principle of least surprise, because users who install a theme shouldn't expect the theme to essentially de-fontify faces, unless that's the stated purpose of the theme--which, for Solarized, it is not.

So, according to Org, macros should stand out from regular text.  But now, in this theme, they do not.  I feel like that's almost violating users' trust (in a very mild way).

Users who want to essentially de-fontify faces should do so in their personal config, or they should install a theme whose purpose is to defontify and "low-contrastify" faces.

IOW, I (politely) object to this, my favorite theme, suddenly imposing what is virtually a semantic change, rather than simply "solarizing" the faces in my Emacs.  It feels like imposing one user's personal preferences on all of the theme's users.

thomasf:
Well, the DEV-GUIDE.md in the root of this repository actually states that we prefer hiding information some times:
> It's sometimes even preferable to hide information by reusing colors rather than creating more visual noise. It's hard to decide (for other people) what to simplify/reduce away but it leads to a better reading experience.
Where should I add this information to make it more prominent?

grettke:
Hi @alphapapa,


> I feel like that's almost violating users' trust

That sounds horrible, truly horrible.

There is no noun or any other word that can follow "violating users' trust"
that changes anything. There is certainly nothing that anybody could say to
me, at least, and I'm probably pretty common in that regard. I hear you and
that sounds horrible. Thank you for sharing that because that gets to the
heart of it. I am sorry here, because that was not my intent at all.

Here is how I felt wrote this:
- I love Org Mode and want the world to enjoy it
- I love Solarized and want the world to enjoy it
- I felt like Org Mode's theme doesn't have the style and color
  options to provide a cohesive experience. It wasn't a focus when they made
  their theme. They couldn't draw from the not just the colors, but the style
  and intent of the style. But Solarized can.
- I felt that the macro needs to stand to stand out, and that bright yellow
  was probably an arbitrary choice and that +1white was a better one.

So I confess to that. No misleading. No violating trust. As you saw, I did put it in a
pull request with pictures so I wasn't being sneaky. I had a really good
intention although it went horribly. I guess that is one of my talents :).

I looked and Solarized covers at least 150 packages so I bet that a lot of
people face the questions of whether or not to do a apples for apples port of
the original color theme to Solarized. For a lot of them there is no concept of the options that Solarized provides, so there are better options available that accentuate the package. For a lot of them I bet that it isn't apples to apples because frankly Solarized is a better theme than the stock
theme for those packages. Most package color themes are probably pretty
random.

At the end of the day though you are right about this patch: I took my
personal preferences and submitted the patch without reading the contribution
guidelines. My intent was innocent though.

So thanks for both of your feedback, it is helping me better understand the right way
to do or not to do Solarized. I look forward to learning more, too.

Thanks and have a great day.

thomasf:
>At the end of the day though you are right about this patch: I took my
personal preferences and submitted the patch without reading the contribution
guidelines. My intent was innocent though.

You did comply with this themes guidelines though even if you maybe didn't read it :)

grettke:
> You did comply with this themes guidelines though even if you maybe didn't read it :)

😅

alphapapa:
Grant, forgive me, I tried to soften my words, but I may have been too harsh.  Instead of saying "violating users' trust" I should have said "violating users' expectations."  After all, we are just talking about colors, here.  It's not like we're introducing security holes.

> Here is how I felt wrote this: ...

Of course, I know your intentions are pure, and I share them; it's only because of that that I spoke up.  In no way were you sneaky or deceptive.  If I hadn't happened to notice this PR, I wouldn't have said anything, and likely would never have noticed.  Maybe I shouldn't have said anything.

In this particular case, I think that Org macros are intended to stand out, because they are code that is run on export rather than plain text.  And I think that users would generally expect that we wouldn't muddle that distinction in a color theme.  Thomas is right that there are only so many colors to go around in this theme, and some editorial discretion is inevitable.  But IMHO this is a case where the color should stand out from plain text, as Org's authors intended.

What if we used a color blended with `solarized-color-blend` that would stand out from plain text but not be as bright as a pure Solarized color?

thomasf:
> What if we used a color blended with solarized-color-blend that would stand out from plain text but not be as bright as a pure Solarized color?

Nah, it's best to avoid non standard colors where possible. 

Maybe adding bold and/or base02 bg would work...

grettke:
@alphapapa Gotcha. Thanks for sharing that

The plan to move forward sounds great.  Can't wait to see it!

PR #309, opened 2018-08-08
Do not derive outline faces from org faces

### LINKED ISSUE #308: Outline faces should not inherit from Org faces
Org is based on Outline -- not the other way around, so we cannot assume that the Org faces are available when someone is using `outline-mode`.

This can be fixed either by inverting the direction of the `:inherit` or by theming the two groups of faces independently.

tarsius:
Org is based on Outline -- not the other way around, so we cannot
assume that the Org faces are available when someone is using
outline-mode.

This can be fixed either by inverting the direction of the :inherit
or by theming the two groups of faces independently.  This commit
does the later.

Additionally this adds a new option `solarized-scale-outline-headlines`,
which is like the existing `solarized-scale-org-headlines`, just for
another package.

Fixes #308.

### DIFF
--- a/solarized.el
+++ b/solarized.el
@@ -112,7 +112,12 @@ Related discussion: https://github.com/bbatsov/solarized-emacs/issues/158"
   :group 'solarized)
 
 (defcustom solarized-scale-org-headlines t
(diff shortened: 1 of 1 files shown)


bbatsov:
Thanks!

PR #312, opened 2018-09-01
Underline current match in Ivy

joonoro:
`ivy-current-match` is too hard to see when using `C-h v` since some of the matches are already highlighted. This change simply underlines the current match, making it easy to distinguish.

Before:
solarized_default
After:
solarized_edited


bbatsov:
Thanks!

NightMachinary:
@bbatsov Any way to disable this? My ivy just appends `>` to its marks, so I don't need the underline. I also like to disable the underline for the modeline:
image

I took a look at the code, but it seemed that I would need to fork it in order to change these?

PR #316, opened 2018-10-30
Theme outline-minor-faces

tarsius:
Unlike `outline-mode`, `outline-minor-mode` does not change
the appearance of headings to look different from comments.

The `outline-minor-faces` package defined faces `outline-minor-N`,
which inherit from the respective `outline-N` faces and arranges
for them to be used in `outline-minor-mode`.

* https://github.com/tarsius/outline-minor-faces
* https://github.com/tarsius/backline might also be of interest. It makes sure that the background color of the heading "extends all the way to the right edge of the window" even when the section body is collapsed.


bbatsov:
Thanks!

PR #327, opened 2019-02-20
Remove org-macro, org-block-begin-line, and org-block-end-line

grettke:
Each violate the principle of lease astonishment with explanations in the commits.


thomasf:
There can be a few instances where it's very unclear where blocks starts or ends without the over/underlines..  I don't remember the specifics but I won't merge this directly.

grettke:
Understood. What about org-macro?

PR #328, opened 2019-02-24
Added support for theming tabbar plugin

ema2159:
Found that tabbar plugin was unthemed so I themed it according to the theme colors,


ema2159:
Currently it looks like this:
tabbar

bbatsov:
Thanks!

PR #331, opened 2019-06-01
Improved Solaire mode colors

ema2159:
Probed Atom's Solarized dark's dark background and added it to the solaire mode faces background.
It looks like this compared to what it originally looked like:
Original
DeepinScreenshot_select-area_20190601110145

With new background:
DeepinScreenshot_select-area_20190601112152

### DIFF
--- a/solarized.el
+++ b/solarized.el
@@ -148,6 +148,7 @@ Alpha should be a float between 0 and 1."
           (variant ,variant)
           (s-base03    "#002b36")
           (s-base02    "#073642")
+          (solaire-dark "#001F27")
           ;; emphasized content
           (s-base01    "#586e75")
           ;; primary content
@@ -198,6 +199,7 @@ Alpha should be a float between 0 and 1."
           (base02 (if (eq variant 'light) s-base2 s-base02))
           (base3 (if (eq variant 'light) s-base03 s-base3))
           (base03 (if (eq variant 'light) s-base3 s-base03))
+          (solaire-base (if (eq variant 'light) s-base2 solaire-dark))
 
           ;; Line drawing color
           ;;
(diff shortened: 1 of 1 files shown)


solarized.el
```diff
@@ -148,6 +148,7 @@ Alpha should be a float between 0 and 1."
           (variant ,variant)
           (s-base03    "#002b36")
           (s-base02    "#073642")
+          (solaire-dark "#001F27")
```
bbatsov: Ideally we should tie this colour somehow to the existing ones, as having an extra colour in the default palette for a single mode looks a bit weird. Guess it should be some variation of `s-base`, right?

ema2159: Hi mr. Batsov! I've already forgotten this one, sorry for the loose PR! The solaire-dark should be just a darkened version of the default color, because in the default palette (Ethan Schoonover's) for the theme it's not considered that there can be darker backgrounds that the one used for editing, so what other editors do is to use a darkened version of that background, and usually they use the color I use in solaire background. How should I rename it though?

thomasf: I don't like adding colors that change the quite delicate contrast ratios of solarized unless it's absolutely necessary. I'm not very fond of changes that are only made for one of the dark/light themes since the idea is that you can change between the theme variants while keeping the same overall look and contrast ratios everywhere.

ema2159: I think that there's also a light variant of that background color, I can add it. Also I think the colors should not present a problem given they're commonly used in every other editor that implements Solarized (i.e. Atom, VS Code...) and given that it is a completely optional customization (it won't show unless the user enables Solaire) I think it is a nice addition.

thomasf: It does look a bit weird to me in general. The non file buffers will get more sharp contrast so they will stand out more while  file buffers generally are the most important buffers and should not be less emphasised than other buffers? In any case since it is for a specific mode I guess it's ok but you don't need to bind the color to a named variable, just add it directly inside the facedefs.

ema2159: The purpose of solaire is to darken non file visiting buffers, but it also has a function to reverse the colors so it's the file visiting buffers  which are dark. Ok, shall I do that? I thought that binding a color like that was considered a bad practice and that everything should be declared in a variable to keep order

bbatsov: You can have some binding only around the solaire faces with some explanation why the extra colour is needed.

ema2159: I think it's better to leave it as it is right now! Introducing new colors would damage the structure of the original theme.

PR #339, opened 2019-09-15
Add transient-separator from transient package

LeandroCardoso:
Add transient-separator from transient package (mainly used by magit)


bbatsov:
You'll have to rebase on top of `master`.

tarsius:
And for Emacs 27 you have to set `:extend` (search the file for examples).

(Also because this is such a thin line and such a similar color, think about whether this is actually necessary at all.)

LeandroCardoso:
New PR replacing this one.

PR #344, opened 2019-11-02
Add CI environment

conao3:
Hi!
I add CI environment.
Your work is only doing `travis enable` to enable CI via travis.


conao3:
There're conflict. I'll fix it.

bbatsov:
Please, use CircleCI or GitHub Actions instead. I've got some serious reservations about the future of Travis and I've replaced it in all of my projects.

conao3:
I remove `.travis.yml`. I don't familiar with CircleCi or GitHub Actions, so you prefer to use it, add these configure yml file on your project.

bbatsov:
I also strongly prefer Buttercup over ert as the test framework - way more expressive syntax, better test runner, etc.

conao3:
use buttercup instead of ert-runner.
there are no tests, but ensure package files are byte-compilable is also important CI merit.

bbatsov:
Btw, enabling Circle is just as easy as Travis - here's a basic reference https://github.com/clojure-emacs/clojure-mode/blob/master/.circleci/config.yml

thomasf:
personally I prefer drone but go with whatever. I don't really see a big win in having CI here.  (afaik) MELPA will just collect and publish files regardless of CI status as long as they compile (?). More importantly the absolute majority of issues here are about things look which is hard to write tests for and impossible to maintains (since it includes hundreds of external packages). I'm not opposed to it, just don't see a big advantage.

conao3:
Certainly, MELPA does not check the status of the CI, but your mentioned situation is only when you ignore the CI status for PR and merge it.
Systematic testing of color themes will be difficult. However, just checking whether it can be `read` by Emacs has the benefit of CI.
This makes maintenance easy and there is no reason to actively deny.

conao3:
There're many conflicts. Close.

PR #346, opened 2019-11-02
Remove turning on rainbow-mode magic comment

conao3:
rainbow-mode is very useful but whether turn on or off
is the decision of the developer.
Pop up execute magic comment every and every time is
very tired.

### DIFF
--- a/solarized.el
+++ b/solarized.el
@@ -2515,7 +2515,6 @@ customize the resulting theme."
 
 ;; Local Variables:
 ;; no-byte-compile: t
-;; eval: (when (fboundp 'rainbow-mode) (rainbow-mode 1))
 ;; indent-tabs-mode: nil
 ;; fill-column: 95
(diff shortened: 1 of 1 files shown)


wyuenho:
Yeah I agree with this change. It isn't very useful in this package anyway as it highlights all the palette colors as if they are HTML colors. `fontify-face-mode` is much more suitable for reviewing the colors on the face names.

thomasf:
you know you can uninstall rainbow mode if you don't like it? :)  It's kind of useful, I don't need to see the actual solarized colors, just some representation of them to quickly get an idea about where colors are used. But sure, it doesnt need to be in the footer either.

conao3:
Yes. I already used it in global. So,
> But sure, it doesnt need to be in the footer either.

I agree with it.

conao3:
Thanks!

PR #348, opened 2019-11-05
color/palette generation improvements

thomasf:
written in Go for now because of more complete color library available than what color.el has.

First steps

- get lightness contrast up a bit.

- undersand what do do about the old -l -d and -hc colors, they should be phased out or changed with this feature

- magit is one of the most complicated things to theme so first steps is to understand what it needs and make sure we are developing towards that goal.


thomasf:
lightness contrasts 

whilie it isnt the only important value, this is a table for the current state ..  all values are lightness (L in LAB),  lc = light value difference between the two preceeding colors.

image

thomasf:
## some initial emerge examples

image


image

image


(blue added as yellow to the theme by accident in the following screenshot)

image

Issue #349, opened 2019-11-05
Wrong number of arguments: (3 . 3), 4


vzaliva:
Using most recent version from MELPA, load-theme solarized-dark (via M-x) gives error:
```
helm-M-x: Wrong number of arguments: (3 . 3), 4
```

vzaliva:
it was working OK last week. Was broken with recent updates.

thomasf:
I just updated from melpa and it works fine. I can't help you, first try to update again to see if it solved the issue.

What is your full emacs version and what does it say if you turn on debugging `(toggle-debug-on-error)` before trying to load the theme?

vzaliva:
Here is a debug stack:

```
Debugger entered--Lisp error: (wrong-number-of-arguments (3 . 3) 4)
  color-rgb-to-hex(0.7631372549019608 0.88 0.6066666666666667 nil)
  apply(color-rgb-to-hex (0.7631372549019608 0.88 0.6066666666666667 nil))
  solarized-color-blend("#ddffdd" "#859900" 0.7)
  (list (quote :background) (solarized-color-blend "#ddffdd" green 0.7))
  (list light-class (list (quote :background) (solarized-color-blend "#ddffdd" green 0.7)))
  (list (list light-class (list (quote :background) (solarized-color-blend "#ddffdd" green 0.7))) (list dark-class (list (quote :background) (solarized-color-blend "#446644" green 0.7))))
  (list (quote diff-refine-added) (list (list light-class (list (quote :background) (solarized-color-blend "#ddffdd" green 0.7))) (list dark-class (list (quote :background) (solarized-color-blend "#446644" green 0.7)))))
  (\` (diff-refine-added (((\, light-class) (:background (\, (solarized-color-blend "#ddffdd" green 0.7)))) ((\, dark-class) (:background (\, (solarized-color-blend "#446644" green 0.7)))))))
[...]

PR #351, opened 2019-11-06
call color-rgb-to-hex with 3 arguments when emacs<26

conao3:
The error is my fault. The error was reported in #349, and this PR
solve it.

solarized-color-blend and solarized-color-blend-rgb might have same
duty, so which one need remove.


thomasf:
thanks

conao3:
thanks!

Issue #352, opened 2019-11-06
child themes does not work


thomasf:
It does not seem that emacs want to redefine a face in a theme after its initially been set. 

This fixes it for me but I'm not sure. Moving the let binding to somewhere where it's actually called for all themes (solarized-definition?) might also work.

```el
diff --git a/solarized-faces.el b/solarized-faces.el
index f2eee33..dd56e62 100644
--- a/solarized-faces.el
+++ b/solarized-faces.el
@@ -12,6 +12,9 @@ customize the resulting theme."
 ;;; Color palette
   (eval
    `(solarized-with-color-variables ',variant ,color-palette
+;;; Child theme Faces
+      ,(when childtheme
+         `(funcall ',childtheme))
 ;;; Theme Faces
       (custom-theme-set-faces
        ',theme-name
@@ -2066,8 +2069,7 @@ customize the resulting theme."
        `(xterm-color-names-bright [,base03 ,orange ,base01 ,base00
                                            ,base0 ,violet ,base1 ,base3]))
 ;;; Setup End
-      ,(when childtheme
-         `(funcall ',childtheme)))))
+      )))
 
 
 (provide 'solarized-faces)
diff --git a/solarized.el b/solarized.el
index 3bb2421..1529232 100644
--- a/solarized.el
+++ b/solarized.el
@@ -417,8 +417,7 @@ If OVERWRITE is non-nil, overwrite theme file if exist."
               `((require 'solarized)
                 (deftheme ,theme-name
                   ,(format "The %s colour theme of Solarized colour theme flavor." theme-name))
-                (let ((custom--inhibit-theme-enable nil))
-                  (solarized-create-theme-with-palette ',variant ',theme-name ',core-palette ',childtheme))
+                (solarized-create-theme-with-palette ',variant ',theme-name ',core-palette ',childtheme)
                 (provide-theme ',theme-name)
                 (provide ',(intern (format "%s-theme" theme-name)))))))
     path))
```

conao3:
First, I think child-theme is not a good idea for the user customization scheme.
Because of the child-theme mechanism existing, solarized cannot enable the `lexical-binding` required for modern packages.

It is better to change the current `solarized-with-color-variables` to `with-solarized-color-variables`, and the upper function should be designed to accept child theme as normal sexp instead of function.
This is a bit painful, but it is an incompatibility necessary to review it for a better design.

thomasf:
[...]

PR #353, opened 2019-11-06
Enable lexical-binding and some code-refactoring

conao3:
Hi!
This PR related #352.  This PR enables lexical-binding, change
child-theme format but I cannot figure out
`custom--inhibit-theme-enable` hack.

So I remain `custom--inhibit-theme-enable` to nil to works well
child-theme feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,12 +18,12 @@ customizations and `deftheme` and does not require the
 ## Installation
 
 Solarized for Emacs is available for installation via the
-[MELPA](http://melpa.org)  `package.el`
-repository. Assuming you've set it up you can install Solarized like this:
+[MELPA](http://melpa.org) using `package.el`.  Assuming you've set it
+up you can install Solarized like this:
 
 `M-x package-install solarized-theme`
 
-This package will install two variants of the theme - `solarized-light-theme`
+This package will install two variants of the theme; `solarized-light-theme`
 and `solarized-dark-theme`. You can load one of the theme variants with `M-x
 load-theme`.
 
@@ -97,6 +97,77 @@ manner as the vertical border.
 ```el
 (setq x-underline-at-descent-line t)
 ```
+
+## Create theme using color palette
+The Solarized Face settings consist of a palette of colors with eight
+accents in addition to the darkest and brightest colors. Recent
+changes allow you to freely create theme files using different
+palettes instead of the Solarized color palette.  It consists of two
+steps: **creating a theme file** and **loading a theme**.
+
+### Creating/Loading Theme Files
+Select the darkest and lightest colors and the eight accents to pass
+to the function. This creates a theme file in `.emacs.d/themes/`. If
+you need to make minor modifications, you can override Face
+individually by specifying a free Sexp as the fourth argument.
+
+Once you have a theme file, you can load it with load-theme, like
+solarized.
+
+```el
+;; inspired vim's jellybeans color-theme
+(solarized-create-theme-with-pallet 'light 'solarized-jellybeans-light
+  '("#202020" "#ffffff"
+    "#ffb964" "#8fbfdc" "#a04040" "#b05080" "#805090" "#fad08a" "#99ad6a" "#8fbfdc"))
+
+(load-theme 'solarized-jellybeans-light t)
+```
+
+```el
+;; inspired emacs's mesa color-theme
+(solarized-create-theme-with-pallet 'light 'solarized-mesa-light
+  '("#000000" "#faf5ee"
+    "#3388dd" "#ac3d1a" "#dd2222" "#8b008b" "#00b7f0" "#1388a2" "#104e8b" "#00688b"))
+
+(load-theme 'solarized-mesa-light t)
+```
+
+```el
+;; inspired emacs's solarized color-theme
+(solarized-create-theme-with-pallet 'light 'solarized-solarized-light
+  '("#002b36" "#fdf6e3"
+    "#b58900" "#cb4b16" "#dc322f" "#d33682" "#6c71c4" "#268bd2" "#2aa198" "#859900"))
+
+(load-theme 'solarized-solarized-light t)
+```
+
+```el
+;; wombat color-theme with misc face definition
+(solarized-create-theme-file 'dark 'solarized-wombat-dark
+  '("#2a2a29" "#f6f3e8"
+    "#e5c06d" "#ddaa6f" "#ffb4ac" "#e5786d" "#834c98" "#a4b5e6" "#7ec98f" "#8ac6f2")
+  '((custom-theme-set-faces
+     theme-name
+     `(default ((,class (:foreground ,(solarized-color-blend base03 base3 0.15 2) :background ,base03))))
+     `(highlight ((,class (:background ,violet))))
+     `(font-lock-builtin-face ((,class (:foreground ,magenta))))
+     `(font-lock-constant-face ((,class (:foreground ,blue))))
+     `(font-lock-comment-face ((,class (:foreground ,base00))))
+     `(mode-line
+       ((,class (:foreground ,base2 :background ,(solarized-color-blend base03 base3 0.85 2)))))
+     `(mode-line-inactive
+       ((,class (:foreground ,base00 :background ,(solarized-color-blend base03 "black" 0.85 2)))))
+     `(mode-line-buffer-id ((,class (:foreground ,base3 :weight bold))))
+     `(minibuffer-prompt ((,class (:foreground ,base1))))
+     `(vertical-border ((,class (:foreground ,base03)))))))
+
+(load-theme 'solarized-wombat-dark t)
+```
+
+### Caution
+If the theme file already exists, `solorized-create-theme-file` does not
+regenerate the file. If the file exists, it can be overwritten by
+setting the fifth argument to `t`.
 
 ## Bugs & Improvements
 
--- a/child-theme-example/my-solarized.el
+++ b/child-theme-example/my-solarized.el
@@ -1,8 +1,7 @@
-(defun my-solarized-theme ()
-  "My solarized child theme."
-  (custom-theme-set-faces
-   theme-name
-   `(default ((,class (:foreground ,red :background ,base03))))
-   ))
+(defvar my-solarized-theme
+  '("My solarized child theme."
+    (custom-theme-set-faces
+     theme-name
+     `(default ((,class (:foreground ,red :background ,base03)))))))
 
 (provide 'my-solarized)
--- a/child-theme-example/themes/my-solarized-dark-theme.el
+++ b/child-theme-example/themes/my-solarized-dark-theme.el
@@ -1,9 +1,12 @@
 (require 'solarized)
 (require 'my-solarized)
+(eval-when-compile
+  (require 'solarized-palettes))
 
 ;; This files needs to be places iside the custom-theme-load-path list
 
 (deftheme my-solarized-dark "The dark variant of the Solarized colour theme")
-(solarized-create-theme 'dark 'my-solarized-dark 'my-solarized-theme)
+(solarized-with-color-variables
+  'dark 'my-solarized-dark solarized-dark-color-palette-alist my-solarized-theme)
 
 (provide-theme 'my-solarized-dark)
--- a/child-theme-example/themes/my-solarized-light-theme.el
+++ b/child-theme-example/themes/my-solarized-light-theme.el
@@ -1,9 +1,12 @@
 (require 'solarized)
 (require 'my-solarized)
+(eval-when-compile
+  (require 'solarized-palettes))
 
 ;; This files needs to be places iside the custom-theme-load-path list
 
 (deftheme my-solarized-light "The light variant of the Solarized colour theme")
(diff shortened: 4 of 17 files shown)


thomasf:
I will check this maybe tomorrow evening, have to do other things for a day or maybe two.

conao3:
OK!

thomasf:
I think we should be more careful about doing things that potentially breaks someone usage.  Maybe we could even move the work to a separate branch not picked up by MELPA for a little while to get a new stable 2.0 version done, what I don't want is to potentially break things too often by not having a clear end goal.  I am not going to merge this because currently it doesn't really fully solve a problem someone actually has (?)

For the short term I will mostly be working on getting rid of all uses of  the accent color -l -d -hc -lc variants usage inside the theme OR clearly define what those colors are and how they should be used. I will try to compile my list of the the best and worst things about how the theme works and what I might expect if it were to have a changed API

conao3:
First, it is wrong to use macros in functions that contain things that you want to change dynamically.
To that end, we're using an dirty `eval`, and in previous versions, it was forced disabling compilation.
Here is some advice from Emacs maintainers.
```elisp
(defmacro dont-compile (&rest body)
  "Like `progn', but the body always runs interpreted (not compiled).
If you think you need this, you're probably making a mistake somewhere."
  (declare (debug t) (indent 0) (obsolete nil "24.4"))
  (list 'eval (list 'quote (if (cdr body) (cons 'progn body) (car body)))))
```
And `lexical-binding nil` is not in line with MELPA's criteria.
Here is the asymmetry of MELPA, MELPA will not be given the opportunity to review once registered packages.
Now it will be difficult to pass the MELPA maintainer review.

My change allows byte-compilation of the entire solarize, no function wrapping macros, and `lexical-binding` can be enabled.
It’s true that I’ve rewritten many things, so it’s only natural to spend time on reviews. However, keeping this branch and repeating other changes is not a good choice.
At some stage, I will be reluctant to resolve conflicts, and this branch will not be maintained and it will never be merged.

bbatsov:
For the protocol - I like the proposed changes. I do feel, however, that all this theme creation machinery has to be documented somewhere as now it's pretty weird. The README doesn't even mention the extra themes shipped with Solarized lately.

thomasf:
Why don't you use the `solarized-` prefix for `with-solarized-color-variables`?

conao3:
Because of Elisp and Lisp conventions.
`solarized-with-color-variables` defines a macro that binds the color palette to the appropriate let variable and executes the `body` in that `let` environment.

A macro that accepts such a body and temporarily executes an S-expression in a different environment has a `with-` prefix to give Lisp programmers a proper understanding.
I believe that we should follow `package-lint` entirely, and using the `with-` prefix is legal under the direction of package-lint.
see https://github.com/purcell/package-lint/pull/146

alphapapa:
@conao3 Steve said on that issue that he does *not* encourage doing this:

> This is fine, but I'm generally a little -1 on allowing these various prefixes, because there are usually reasonable possible alternatives, and these prefixes aren't often used anyway: I wouldn't want to actively encourage them.

What benefit is there to changing the macro's name from `solarized-with-color-variables` to `with-solarized-color-variables`?  It's already correct according to MELPA and Elisp guidelines: it begins with the package prefix (which is the most important, because it aids discovery of a package's functions and macros in Emacs help facilities), then the `with` prefix, then the rest of the name.  The docstring is to explain what it actually does.

> Here is the asymmetry of MELPA, MELPA will not be given the opportunity to review once registered packages.  Now it will be difficult to pass the MELPA maintainer review.

MELPA has thousands of packages that were added before its requirements were more strictly enforced.  It is not necessary, nor even necessarily helpful, to make such changes for the sake of making such changes.

Please do not make these kinds of churning changes that do not benefit users.

> At some stage, I will be reluctant to resolve conflicts, and this branch will not be maintained and it will never be merged.

This sounds like you're trying to leverage code you wrote which has changes that are desired to force through changes that are not desired.  "Merge all of the changes or I will let them rot" is not a good attitude.

conao3:
What is the **user benefit** that you claim?
If you only want to use the code at a certain point forever, you can specify a commit hash in Cask and continue to use it personally. It is your choice and do as you please.

There're no meanings more than shortening the life of the Emacs community by stopping people from writing better code by interfering with them.

It is important to follow the package-lint warning.
This simply indicates the maintenance status of the package, and a package that has been left with dozens of warnings gives a dangerous signal about the maintenance status.

alphapapa:
You've said several times now, in this and other issues, that I should not criticize proposals, because if I do, people will stop working on the project, and it will die.  Therefore, it is implied, all proposals are a net positive, and your proposals must be accepted as-is.  

It's impossible to reason about technical issues by starting from that premise.  It is a very hostile attitude to insist that your proposals be accepted or you will stop working on the project.

conao3:
When did I insist that it should be accepted as-is?
As a user of this package, I just make suggestions to improve the package.
And it is rare for a proposal to be accepted as it is, and in many cases, with the maintainer advice try to modify the proposal and cooperate with it.

You just block it. There is only one point. Don't change the code.
This only prevents the package from continuing to evolve, and it will be also detrimental to you if this package is not updated.

alphapapa:
We don't seem to be communicating very well about these issues.  I don't know why.  But I won't bother you about them anymore.  I look forward to the planned improvements.

thomasf:
I'm not a big fan of with- . Doing a quick apropos for after force loading my 500+ packages elpa revelaes that it's probably not a widely used idiom. packatename-with-... is much more common and what I personally would expect from anything not in the standard library. Just because a linter rule checker alllows for something does no always imply that it's the right thing to do.
When i personally don't remember exactly what some package function  I always start by doing `C-h f` `packagename-` `COMPLETE`  to quickly see a list.  I just expect packages to stay in their prefixes because it's a sane overall method of organizing things.

conao3:
OK. Fix that.

thomasf:
I am holding off doing any more commits to master btw until this is either merged or decided against.  It certainly looks like a good step on the path towards a better package.

I so much would like the child theme functionality just be removed and reintroduced later in a better way that actually works but it still possible to use it now even in its broken stage and don't want remove it without having an alternative.

Even if it's not overwriting already set faces correctly it still works for my custom-set-variables:

```el
(defun my-solarized-theme ()
  "My solarized child theme."

  (custom-theme-set-faces
   theme-name


   ;;                                  _
   ;;  _ __  ___ _ _ ___ ___ _ _  __ _| |
   ;; | '_ \/ -_) '_(_-</ _ \ ' \/ _` | |
   ;; | .__/\___|_| /__/\___/_||_\__,_|_|
   ;; |_|
   ;; what other people might not want


   `(rainbow-delimiters-depth-1-face ((,class (:foreground ,base02))))

   )

  (custom-theme-set-variables
   theme-name

   `(org-todo-keyword-faces
     (quote (("TODO" :foreground ,cyan :weight bold :slant italic :underline nil)
             ("NEXT" :foreground ,blue :weight bold :slant italic :underline nil)
             ("ACTIVE" :foreground ,blue :weight bold :slant italic :underline nil)
             ("DONE" :foreground ,base01 :weight bold :slant italic :underline nil)
             ("WAITING" :foreground ,yellow :weight bold :slant italic :underline nil)
             ("HOLD" :foreground ,yellow :weight bold :slant italic :underline nil)
             ("NOTE" :foreground ,green :weight bold :slant italic :underline nil)
             ("SOMEDAY" :foreground ,base01 :weight bold :slant italic :underline nil)
             ("CANCELLED" :foreground ,base01 :weight bold :slant italic :underline nil)
             ("PHONE" :foreground ,blue :weight bold :slant italicg :underline nil))))

   `(jump-char-lazy-highlight-face nil)
   `(org-tag-faces
     '(("@home"  :foreground ,green)
       ("@office" :foreground ,blue)
       ("@errand" :foreground ,cyan)
       ("@travel" :foreground ,violet)
       ("task"  :foreground ,yellow)
       ("bug"  :foreground ,yellow)
       ("enhancement"  :foreground ,yellow)
       ("work" :slant italic :foreground ,base01)
       ("personal" :slant italic :foreground ,base01)
       ("23c" :slant italic :background ,base02)
       ("bulk" :foreground ,base01)
       ("hold" :foreground ,green)
       ("note" :foreground ,base01)
       ("waiting" :foreground ,orange)
       ("cancelled" :foreground ,green  )
       ("flagged" :foreground ,red))

     ;; custom-set-variables end
     )))
```

conao3:
I have not removed the child-theme functionality. The explanation is included in another PR(#355 ).
In fact, it's important to create a theme file and how to run the theme.
I changed both to make lexical-binding work, but with a simple change you can use child-theme under a new mechanism.

In your example, you can simply delete the function name, arguments, and docstring and pass it as a list to the theme creation function.
In these situations, you should use Lisp's **homoiconic** rather than functions, and this method works well with lexical-binding.

thomasf:
Or may

> The README doesn't even mention the extra themes shipped with Solarized lately.

Yeah, my bad. I will add that to the readme after this and the other readme change has been merged. Need some time to think about how to formulate that text in an approachable way anyways.

thomasf:
> I have not removed the child-theme functionality. The explanation is included in another PR(#355 ).
I'm not saying you have. It's probably best to repair it towards something that is more like it worked (I guess before emacs theme system changed something internally) now and postpone any potential structural changes to customization to a later date.

> In fact, it's important to create a theme file and how to run the theme.
Yeah, I just remembered that function, I have just tried it briefly and not actually used it yet. 

We are still relying on `custom--inhibit-theme-enable` which from what I can tell will in general sitirr up how the emacs theme system applies themes over eachother?

Did you have any thought about the solution of just running it before the themes own custom-set-faces instead? I don't know if custom-set-variables has the same "write protection" rules though, if thats not the case we are just exchanging one problem for another by changing the order.

conao3:
I found child-theme does not work well. Please, some time to investigate.

conao3:
Issue solved.
And I think `custom--inhibit-theme-enable` related issue should be improved another PR or commits.

thomasf:
I will try actually using this branch later in the week end to see if I find any problems

conao3:
OK. Please test.

thomasf:
please don't remove the child themes from the repository, I need those to be available when using the `make dev` environment to be sure that they are always working as expected. I think it's even a good idea to set an very hard color in them like making the foreground color #ff0000 or something so it's always clear that they actually do work.

conao3:
Revert remove child-theme example files commit. and fix those files to fit new API.

thomasf:
solarized-wombat-dark-theme.el does not work since it now creates a theme file instead of actually being a theme (?).

I think you can move it to the minimal-init test env for now, I can take care of loading it into the test env after merge.

I'm thinking that we maybe should add a changelog/migration file that just explains what people should do if they have an old child theme and what to use it now. Just something that people that don't know elisp can follow.

conao3:
Thanks for your point. I fix that.
BTW, I have thought these `sample` theme files should remove. But the discussion is another issue.

thomasf:
Yeah, the extra theme files can be moved into their own elpa package also build from this repository or something. I still want them to be installable but since they cause confusion about what they are it's probably for the best to not have them in the main package.

I'm probably merging this within a hour or so now.  Maybe add an upgrade.guide.md in the root aftewards. I'm thinking along the lines of a format something like this (just placeholder content atm)

# solarized theme upgrade guide


## 2019-11-17

If you are using child themes you need to modify your child themes like this

### before

**my-solarized.el**

```el
(require 'solarized)

(defun my-solarized-theme ()
  "My solarized child theme."
  (custom-theme-set-faces
   theme-name
   `(rainbow-delimiters-depth-1-face ((,class (:foreground ,base02)))))
  (custom-theme-set-variables
   theme-name
   `(org-todo-keyword-faces
     (quote (("TODO" :foreground ,cyan :weight bold :slant italic :underline nil))))))

(provide 'my-solarized)
```

**my-solarized-dark-theme.el**
```el
(require 'solarized)
(require 'my-solarized)

(deftheme my-solarized-dark "The light variant of the Solarized colour theme")
(create-solarized-theme 'dark 'my-solarized-dark 'my-solarized-theme)

(provide-theme 'my-solarized-dark)
```


### after

**my-solarized.el**
```
(require 'solarized)

(setq my-solarized-childtheme  '(
    (custom-theme-set-faces
       theme-name
       `(rainbow-delimiters-depth-1-face ((,class (:foreground ,base02)))))
      (custom-theme-set-variables
       theme-name
       `(org-todo-keyword-faces
         (quote (("TODO" :foreground ,cyan :weight bold :slant italic :underline nil))))))
    )
)
(provide 'my-solarized)
```

**my-solarized-dark-theme.el**

```el
...
```

conao3:
LGTM.

thomasf:
I would prefer a single commit to merge.. 

I think this might be one of the paths of least resistance to create a rebased merge out of these changes: 

I can do it myself but then you won't be the author in git.

```sh
git merge -Xignore-space-change master
git reset master
git add -A
git commit -m"...."
```

conao3:
rebased.

conao3:
Thanks!

thomasf:
the readme update references wrong function names

PR #357, opened 2019-11-09
Move magit-diff-* faces where they belong

tarsius:
Also order solarized-create-color-palette consistently.


thomasf:
doing this as a part of #360

PR #364, opened 2019-11-17
Fix file local variables

conao3:
Since #353 merged, `solarized.el` could be byte-compile.
I remove the needless `no-byte-compile t` configure and
`indent-tabs-mode nil` in every emacs-lisp files.

----

#


conao3:
Thanks!

Issue #366, opened 2019-11-18
solarized-with-color-variables is broken


blaenk:
I contributed this macro way back. From what I gather there are some major changes/refactorings going on, but just in case it was a mistake, `solarized-with-color-variables` appears to be broken.

Its intention is to evaluate the macro's body within a `let` environment with the color variables bound, to allow the colors to be used for anything (such as `setq`'s or whatever), but this is no longer possible. ~~The docstring still references a `BODY` parameter but it no longer exists in its function definition. Now this macro seems to effect theme changes through `custom-theme-set-faces`.~~

I think the general idea of being able to access the colors without side-effects has proven to be very useful, whether it be by this macro or some other way.

conao3:
Please show me *broken* example. Then I can investigate it.

conao3:
I understand. Use *child-theme* feature instead. Please see https://github.com/bbatsov/solarized-emacs/blob/master/UPGRADE-GUIDE.md

blaenk:
Thanks @conao3

PR #367, opened 2019-11-18
fix BODY to solarized-definition

conao3:
fix remained `BODY` variable reference to `solarized-definition`.


conao3:
This issue reported at #366.

conao3:
Thanks!

PR #359, opened 2019-11-11
Explicitly extend many faces beyond eol for Emacs 27

tarsius:
This is *not* intended to be merged.  Eventually I will redo this on top of the recent changes on `master`.

At this time the main purpose of this pull-request is to show Emacs developers what kind of changes will have to be made to virtually every theme if they stick to the plan of adding an `:extend` face attribute that defaults to `nil`.

PR #371, opened 2019-11-20
Explicitly extend many faces beyond eol for Emacs 27 (version 2)

tarsius:
This replaces #359.

This doesn't fix all affected faces, but it at least makes new solarized usable with new emacs, at least for me.

### DIFF
--- a/solarized-faces.el
+++ b/solarized-faces.el
@@ -72,8 +72,13 @@
                                 :background ,s-mode-line-inactive-bg
                                 :box (:line-width 1 :color ,s-mode-line-inactive-bg
                                                   :style unspecified)))))
-     `(region ((,class (:foreground ,base03 :background ,base1))))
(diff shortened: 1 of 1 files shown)


thomasf:
By emacs 27 do you mean the current master branch of the Emacs repository?

tarsius:
Yes.

thomasf:
Just for reference if anyone is wondering.. 

enacs 27 without this patch 

image


emacs 26

image

thomasf:
emacs 26 keeps working as before so I see no harm in merging this for people who runs pre release version of emacs.

tarsius:
Since https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=58fb4c3e68a4a42ad491d0fa2c084e5c39942e2b this isn't necessary anymore and should be reverted.

PR #372, opened 2019-11-26
Merge *color-blend-rgb to *color-blend

conao3:
This issue found at #351.
`solarized-color-blend` is used in `solalized-faces` many times but `solarized-color-blend-rgb` is not used.

And `(solarized-color-blend-rgb "red" "green" 0.3)` occur error with `void variables`. This related enabling `lexical variables` maybe.

My suggestions `solarized-color-blend-rgb` alias to `solarized-color-blend`. Or just remove is also good for me.

### DIFF
--- a/solarized.el
+++ b/solarized.el
@@ -114,25 +114,6 @@ Related discussion: https://github.com/bbatsov/solarized-emacs/issues/158"
 
 ;;; Utilities
 
-;;;###autoload
-(defun solarized-color-blend-rgb (color1 color2 alpha &optional digits-per-component)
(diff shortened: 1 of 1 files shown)


conao3:
Resolve conflict.

thomasf:
It is not used at all in the theme since I renamed the old one to -rgb when adding the cie blending function.  I don't think that anyone has started using that function so the alias can be removed as well.

conao3:
Yes, I think so too.

conao3:
Thanks!

PR #373, opened 2019-11-26
Remove solarized-color-lightness

conao3:
`solarized-color-lightness` defined with argument `color` but references `hexrgb`in function...
And this function is not autoloaded and not used so my proposal is just remove this function.

### DIFF
--- a/solarized.el
+++ b/solarized.el
@@ -133,10 +133,6 @@ use the latter if you need a 24-bit specification of a color."
         (apply 'color-rgb-to-hex `(,@args))
       (apply 'color-rgb-to-hex `(,@args ,digits-per-component)))))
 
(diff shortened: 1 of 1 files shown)


thomasf:
Yeah, I had started to port the stuff needed to rewrite the Go color palette code in elisp but it turned out that the results of `color.el` rgb-to-lab and the `colorful` Go package results in (slightly) different values being produced so I shelved it for now. Looking at how those two are implemented they even use slightly different constants to do the conversion so I just aborted and probably left it broken because I never used it. The hope was that the results would match 100% so that I could use the palette output from the Go tool to assert that the output would stay the same.

We can remove the lightness function for now until I maybe take on the porting again later.

conao3:
Thanks!

PR #376, opened 2019-12-02
Remove dash dependency

conao3:
Hi! I found this package depend dash, but dash function is not used anywhere...?

So I remove dash dependency.  And it is good package no depends such a wrapper library.

### DIFF
--- a/solarized.el
+++ b/solarized.el
@@ -24,7 +24,6 @@
 
 ;;; Code:
 
-(require 'dash)
 (require 'color)
 (require 'solarized-faces)
 


bbatsov:
Thanks!

conao3:
Thanks!

thomasf:
dash is used inside the faces elisp file IIRC

thomasf:
here is one instance

https://github.com/bbatsov/solarized-emacs/blob/master/solarized-faces.el#L1992

conao3:
I see solarized.el only, I overlooked that place! I fix that and send PR.

PR #377, opened 2019-12-02
Add autoload for external API functions

conao3:
Hi! I add autoload for external API functions.

### DIFF
--- a/solarized.el
+++ b/solarized.el
@@ -233,6 +233,7 @@ The Returned color-palette has the same format as `solarized-color-palette'"
       (green-2fg   . ,(solarized-color-blend brightest-base green   0.45 2)))))
 
 ;;; Setup Start
+;;;###autoload
 (defmacro solarized-with-color-variables (variant theme-name color-palette &optional childtheme-sexp)
   "Eval `solarized-definition' in solarized COLOR-PALETTE for THEME-NAME.
 VARIANT is 'dark or 'light.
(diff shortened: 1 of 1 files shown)


bbatsov:
API functions should not be autoloaded - only user commands are meant to be autoloaded. If you're using an API you're supposed to require its library.

conao3:
So why does `solarized-create-color-palette` have autoload cookies?
The function I added a cookie has a function similar to `solarized-create-color-palette`, but rather calls it.
If there is an autoload cookie, the user can omit an explicit `require`.

conao3:
Sorry I use `external API` as `not internal`. These functions I added autoload cookie documented and the user should use them.

bbatsov:
> So why does solarized-create-color-palette have autoload cookies?

Because someone made a mistake. :-) We just have to remove all autoloads. 

> If there is an autoload cookie, the user can omit an explicit require.

As I said - autoloads are not intended to be used in this manner and we should not abuse them.

conao3:
In my understanding, `autoload` should be added all `entry point` function.
`Entry point` is when not loaded the whole package but activated it.

The user uses `solarized-create-theme-file` then `load-theme` that file.
Currently, the user needed `(require 'solarized)` before them but if there are `autoload` cookies, he can omit explicit `require`.

thomasf:
The emacs manual says:

The most common items to autoload are the **interactive entry** points to a library

thomasf:
I think that just using the `###autoload` thing will always wrap a function inside an loading proxy thats interactive but that maybe has changed since I last fiddled with those things (a few emacs major versions ago).

conao3:
Yes, **common**.
But there are entry-pointable functions if the function is non-interactive. You know `use-package` or `bind-key`, don't you?. They are non-interactive but because they are entry point functions, added `autoload` cookie.

I think to remove `autoload` cookie for `solarized-color-blend` too, but `solarized-create-theme-file` etc should be specified autoloaded.

thomasf:
We won't add more autoload cookies, just require the features if you need them elsewhere. Already present autoloads stays for now to stay backwards compatible.

PR #378, opened 2019-12-02
Return raw color if alpha is 1 or 0

conao3:
Hi! I found `solarized-color-blend` has a calculate errors.

``` emacs-lisp
(let ((darkest-base "#3F3F3F")
      (brightest-base "#FFFFFD"))
  (solarized-color-blend darkest-base brightest-base 1.00 2))
;;=> "#3e3e3e"

(let ((darkest-base "#3F3F3F")
      (brightest-base "#FFFFFD"))
  (solarized-color-blend darkest-base brightest-base 0.00 2))
;;=> "#fefefc"
```

So, I fix the function should return raw color if `alpha` is 1 or 0.

### DIFF
--- a/solarized.el
+++ b/solarized.el
@@ -131,16 +131,20 @@ Alpha should be a float between 0 and 1.
 
 Optional argument DIGITS-PER-COMPONENT can be either 4 (the default) or 2;
 use the latter if you need a 24-bit specification of a color."
-  (let ((args (mapcar 'color-clamp
-                      (apply 'color-lab-to-srgb
-                             (solarized-color-clamp-lab
-                              (cl-mapcar
-                               (lambda (v1 v2) (+ v1 (* alpha (- v2 v1))))
-                               (apply 'color-srgb-to-lab (color-name-to-rgb color2))
-                               (apply 'color-srgb-to-lab (color-name-to-rgb color1))))))))
-    (if (version< emacs-version "26")
-        (apply 'color-rgb-to-hex `(,@args))
-      (apply 'color-rgb-to-hex `(,@args ,digits-per-component)))))
+  (cond
+   ((= 1 alpha) color1)
+   ((= 0 alpha) color2)
+   (t
+    (let ((args (mapcar 'color-clamp
+                        (apply 'color-lab-to-srgb
+                               (solarized-color-clamp-lab
+                                (cl-mapcar
+                                 (lambda (v1 v2) (+ v1 (* alpha (- v2 v1))))
+                                 (apply 'color-srgb-to-lab (color-name-to-rgb color2))
+                                 (apply 'color-srgb-to-lab (color-name-to-rgb color1))))))))
+      (if (version< emacs-version "26")
+          (apply 'color-rgb-to-hex `(,@args))
+        (apply 'color-rgb-to-hex `(,@args ,digits-per-component)))))))
 
 ;;;###autoload
 (defun solarized-create-color-palette (core-palette)


thomasf:
I wonder where the error is, probably loss of precision somewhere but it's a bit suspect that it's exactly one below expected value everywhere.

thomasf:
The Go colorful package does not have this problem

```go
func main() {
	c1, err := colorful.Hex("#3F3F3F")
	if err != nil {
		panic(err)
	}
	c2, err := colorful.Hex("#FFFFFD")
	if err != nil {
		panic(err)
	}
	fmt.Println(c1.BlendLab(c2, 0.00).Clamped().Hex())
	fmt.Println(c1.BlendLab(c2, 1.00).Clamped().Hex())
}
```

output
```
#3f3f3f
#fffffd
```

It uses a struct with three float64 R,G,B between 0-1 to represent a color internally.

Maybe it's worth looking at why we don't get the right output, the go implementation of blend lab doing more or less the same thing as ours:

```go
// BlendLab blends two colors in the L*a*b* color-space, which should result in a smoother blend.
// t == 0 results in c1, t == 1 results in c2
func (c1 Color) BlendLab(c2 Color, t float64) Color {
    l1, a1, b1 := c1.Lab()
    l2, a2, b2 := c2.Lab()
    return Lab(l1 + t*(l2 - l1),
               a1 + t*(a2 - a1),
               b1 + t*(b2 - b1))
}
```

thomasf:
only converting from and back color spaces shows as expected the same issue

```el
(message "%S" (apply 'color-rgb-to-hex
                     (append
                      (apply 'color-lab-to-srgb
                             (apply 'color-srgb-to-lab
                                    (color-name-to-rgb darkest-base)))
                      '(2))))
```


not only once either, green consistently goes down by one each conversion cycle until it reaches zero.

```el
(let ((c "#FFFFFD") )
  (--dotimes 255
    (setq c (apply 'color-rgb-to-hex
                     (append
                      (apply 'color-lab-to-srgb
                             (apply 'color-srgb-to-lab
                                    (color-name-to-rgb c)))
                      '(2))))
    (message "%S" c))))
```

thomasf:
in color.el there seems to be an interesting value 

```el

(defun color-srgb-to-xyz (red green blue)
  "Convert RED GREEN BLUE colors from the sRGB color space to CIE XYZ.
RED, GREEN and BLUE should be between 0.0 and 1.0, inclusive."
  (let ((r (if (<= red 0.04045)
               (/ red 12.95)
             (expt (/ (+ red 0.055) 1.055) 2.4)))
        (g (if (<= green 0.04045)
               (/ green 12.95)
             (expt (/ (+ green 0.055) 1.055) 2.4)))
        (b (if (<= blue 0.04045)
               (/ blue 12.95)
             (expt (/ (+ blue 0.055) 1.055) 2.4))))
    (list (+ (* 0.4124564 r) (* 0.3575761 g) (* 0.1804375 b))
          (+ (* 0.21266729 r) (* 0.7151522 g) (* 0.0721750 b))
          (+ (* 0.0193339 r) (* 0.1191920 g) (* 0.9503041 b)))))

```

0.21266729 seems to maybe have one too many 6's. Google for 0.21266729 and you only find color.el, 0.2126729 OTOH brings up a lot of color conversion packages.. However, just chaning that value in color.el makes both red and green to start dropping at conversion.

thomasf:
It probably is a loss of precision issue at the core.. 

I have just been using the go colorful package + google/wikipedia as a reference when looking for things that might be off. Adding rounding to color-rgb-to-hex probably is what actually will solve the issue.

full patch

```diff
diff --git a/lisp/color.el b/lisp/color.el
index 258acbe405..cc478a7525 100644
--- a/lisp/color.el
+++ b/lisp/color.el
@@ -63,7 +63,7 @@ color-rgb-to-hex
   (or digits-per-component (setq digits-per-component 4))
   (let* ((maxval (if (= digits-per-component 2) 255 65535))
          (fmt (if (= digits-per-component 2) "#%02x%02x%02x" "#%04x%04x%04x")))
-    (format fmt (* red maxval) (* green maxval) (* blue maxval))))
+    (format fmt (+ 0.5 (* red maxval)) (+ 0.5 (* green maxval)) (+ 0.5 (* blue maxval)))))
 
 (defun color-complement (color-name)
   "Return the color that is the complement of COLOR-NAME.
@@ -187,16 +187,16 @@ color-srgb-to-xyz
   "Convert RED GREEN BLUE colors from the sRGB color space to CIE XYZ.
 RED, GREEN and BLUE should be between 0.0 and 1.0, inclusive."
   (let ((r (if (<= red 0.04045)
-               (/ red 12.95)
+               (/ red 12.92)
              (expt (/ (+ red 0.055) 1.055) 2.4)))
         (g (if (<= green 0.04045)
-               (/ green 12.95)
+               (/ green 12.92)
              (expt (/ (+ green 0.055) 1.055) 2.4)))
         (b (if (<= blue 0.04045)
-               (/ blue 12.95)
+               (/ blue 12.92)
              (expt (/ (+ blue 0.055) 1.055) 2.4))))
     (list (+ (* 0.4124564 r) (* 0.3575761 g) (* 0.1804375 b))
-          (+ (* 0.21266729 r) (* 0.7151522 g) (* 0.0721750 b))
+          (+ (* 0.2126729 r) (* 0.7151522 g) (* 0.0721750 b))
           (+ (* 0.0193339 r) (* 0.1191920 g) (* 0.9503041 b)))))
 
 (defun color-xyz-to-srgb (X Y Z)
```

thomasf:
color differences before/after when dumping all usage of the blend function when loading solarized light.

Some light shades are noticeably different but nothing that should cause any issues.

image

yamashita-optarc:
Great, thanks!
And I'm glad about your health. take care, and let's continue to enjoy Emacs!

conao3:
Great, thanks!
And I'm glad about your health. take care, and let's continue to enjoy Emacs!

PR #379, opened 2019-12-03
Use mapcar instead of --map

conao3:
Re #376. Use standard function instead of dash.

### DIFF
--- a/solarized-faces.el
+++ b/solarized-faces.el
@@ -1989,8 +1989,9 @@
 ;;;;; highlight-symbol
      `(highlight-symbol-foreground-color ,base1)
      `(highlight-symbol-colors
-       (--map (solarized-color-blend it ,base03 0.25)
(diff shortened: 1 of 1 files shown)


solarized-faces.el
```diff
@@ -1989,8 +1989,9 @@
 ;;;;; highlight-symbol
      `(highlight-symbol-foreground-color ,base1)
      `(highlight-symbol-colors
-       (--map (solarized-color-blend it ,base03 0.25)
-              '(,yellow ,cyan ,red ,violet ,green ,orange ,blue)))
+       ,(mapcar
+         (lambda (it) (solarized-color-blend it ,base03 0.25))
```
bbatsov: I'd use `color` instead of the generic `it`.

bbatsov:
Thanks!

conao3:
Your welcome!

PR #382, opened 2019-12-08
Rename solarized-faces.el to solarized-definitions.el

conao3:
Hi!
I think solarized-faces.el should rename to solarized-definitsions.el

Here is solarized-faces header one-line description.

``` emacs-lisp
the faces definitions for solarized theme
```

And the file define `solarized-definition` variable. So this rename is reasonable for me.


thomasf:
I don't think its a good enough reason to possibley make merging more annoying for people who has forks.

conao3:
Well, ok.

PR #385, opened 2019-12-08
Rename DEV-GUIDE.md -> CONTRIBUTING.md

conao3:
For common agreement, developers notice `CONTRIBUTING` rather than `DEV-GUIDE`.


thomasf:
nah, I'm fine with the name.  CONTRIBUTING is more often about practicalities of development workflow, not documentation. We also create a contributing document that refers to DEV-GUIDE if it's needed

conao3:
Well, OK.

thomasf:
You just get the risk that people don't read the CONTRIBUTING file instead because they assume it doesn't contain actual documentation. I think it just moves the problem

conao3:
I cannot understand your comments.

thomasf:
What I mean is that a lot of people don't expect the CONTRIBUTING file to contain actual project documentation which isn't written elsewhere so they might not read it and then we have the same kind of problem as people not reading DEV-GUIDE.

PR #388, opened 2019-12-12
Add space between close paren and open paren

conao3:
Hi! I found nospace at between close paren and open paren. Fix it.


conao3:
Thanks!

PR #398, opened 2020-04-25
Update elfeed faces

mallt:
This PR updates the elfeed faces.

Before:
screenshot-elfeed-solarized-before

After:
screenshot-elfeed-solarized-after


thomasf:
The current looks are chosen so that the title stands out the most because it’s the most important information of a line.

thomasf:
I’m not sure that adding more colors really contributes to anything important

mallt:
Ok no problem, thanks for checking!

PR #403, opened 2020-08-17
Add recent debbugs-gnu-stale-* faces

skangas:
Please add these faces recently added to the debbugs-gnu package in GNU ELPA.


thomasf:
please provide screenshots, did you not see the pull request template? https://github.com/bbatsov/solarized-emacs/blob/master/.github/PULL_REQUEST_TEMPLATE.md

thomasf:
My immediate reflection is if it's possible to avoid using the fg/bg colors.. As I read the colors (not knowing how debbugs applies them) stale-2-5 will stand out more than any other status.. I don't know what a stale bug is but to me it sounds like something that maybe should stand out less that other statuses.

thomasf:
Adding these for now. Not sure they are the right choice though.

PR #410, opened 2021-01-08
define-obsolete-function-alias added required 3rd parameter

wlharvey4:
The function `define-obsolete-function-alias` changed its optional third parameter into a required parameter, which is a string indicating when or at which version the old function became obsolete.  This change simply adds the required third parameter using "26".  Without this change, `solarized.el` creates an error which prevents the theme from being loaded, but also prevents any other theme from being loaded via the `customize-themes' function.

-----------------

- [x] The commits are consistent with our [contribution guidelines](../blob/master/CONTRIBUTING.md)

Thanks!

### DIFF
--- a/solarized.el
+++ b/solarized.el
@@ -409,7 +409,7 @@ If OVERWRITE is non-nil, overwrite theme file if exist."
   (let ((color-palette (solarized-create-color-palette core-palette)))
     (apply #'solarized-create-theme-file (list variant theme-name color-palette childtheme-sexp overwrite))))
 
-(define-obsolete-function-alias 'create-solarized-theme-file 'solarized-create-theme-file)
+(define-obsolete-function-alias 'create-solarized-theme-file 'solarized-create-theme-file "1.3.0")
 
 ;;; Footer
 


solarized.el
```diff
@@ -409,7 +409,7 @@ If OVERWRITE is non-nil, overwrite theme file if exist."
   (let ((color-palette (solarized-create-color-palette core-palette)))
     (apply #'solarized-create-theme-file (list variant theme-name color-palette childtheme-sexp overwrite))))
 
-(define-obsolete-function-alias 'create-solarized-theme-file 'solarized-create-theme-file)
+(define-obsolete-function-alias 'create-solarized-theme-file 'solarized-create-theme-file "26")
```
bbatsov, line 412: This should be the version of solarized that introduced the deprecation.

wlharvey4:
I have corrected that version number to 1.3.0, from `48290fc53337d3c92acc1c8111aed07222f0ee51` added on 11/3/2019.

bbatsov:
Thanks!

PR #413, opened 2021-01-25
add support for orderless

ssnnoo:
This adds support for the [orderless](https://github.com/oantolin/orderless) completion style.

current

new

-----------------

Thanks!

### DIFF
--- a/solarized-faces.el
+++ b/solarized-faces.el
@@ -1366,7 +1366,11 @@
      `(notmuch-tree-match-date-face ((,class (:foreground ,yellow))))
      `(notmuch-tree-match-tag-face ((,class (:foreground ,cyan))))
      `(notmuch-tree-no-match-face ((,class (:inherit font-lock-comment-face))))
-
+;;;;; orderless
+     `(orderless-match-face-0 ((,class (:foreground ,yellow))))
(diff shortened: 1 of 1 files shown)


thomasf:
I'm not fully sure about the colors.. In the screenshot the blue kind of pushed out the single green color coming directly afterwards and make it less noticeable.   I am thinking that to avoid color mess only using 2 alternating colors would be enough since orderless cycles them anyways?

In any case I am merging this because regardless if it can be improved it's better than what we had before which is nothing.

ssnnoo:
Thanks for merging it! But good point. I’ll play around bit and see if I can improve it. In practice one hardly ever uses more than three different narrowing groups.

thomasf:
I would maybe begin with magenta since it's generally the most visible color and I at least tend to draw my attention to it the most so it's a good color to starts a series of multiple colors off with. 

I think the most common colors for various kinds of search highlights in the theme are magenta, yellow and blue, it depends a little bit on context. It's a good start that all of them are used here.

ssnnoo:
Ah, I started with yellow because this is currently used for ivy and thought this would give consistency. But much better to start with a brighter colour. I will prepare a pull request later today.

PR #417, opened 2021-07-06
Improve handling of auto-highlight-symbol after upstream changes

madand:
This commit accommodates to the following upstream changes:
* `ahs-plugin-defalt-face` was renamed to `ahs-plugin-default-face`
* Several new *-unfocused faces were introduced (e.g. `ahs-face-unfocsed`).

All the mentioned faces are themed exactly the same way the `ahs-face` is.

Please note that the upstream changes to `auto-highlight-symbol` affect Spacemacs (see  syl20bnr/spacemacs#14880) users who happen to use solarized theme. Thus please consider merging this at your earliest opportunity.

Before:
solarized-theme-before

After:
solarized-theme-after

### DIFF
--- a/solarized-faces.el
+++ b/solarized-faces.el
@@ -325,10 +325,14 @@
 ;;;;; auto-highlight-symbol
      `(ahs-definition-face ((,class (:foreground ,magenta :background unspecified
                                                  :slant normal))))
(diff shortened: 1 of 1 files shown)


thomasf:
this looks like the right choice, the same face everywhere is probably for the best.

PR #425, opened 2021-08-22
Add selenized pallets

nmccarty:
This adds support for all four variants of the [selenized color pallets](https://github.com/jan-warchol/selenized), a moderate redesign of solarized.

Selenized Dark: 
selenized-dark-emacs
Selenized Light:
selenized-light-emacs
Selenized Black:
selenized-black-emacs
Selenized White:
selenized-white-emacs


thomasf:
Maybe the strong fg/bg colors could be tweaked a bit for this palette. It is readable as it is now but maybe just slightly too green foreground look.

image

nmccarty:
I'm not going to disagree with you, but I am moderately colorblind so I'm not qualified to make that determination.

PR #435, opened 2022-08-19
Use extra-light instead of thin for line numbers

real-or-random:
#305 was a great but thin is *too* thin for fonts which support many
weights. This replaces thin by extra-light, which is still visibly
different from the normal weight but also still legible.

In fact, it's so thin, it's not even documented, see
https://github.com/emacs-mirror/emacs/blob/master/src/font.c#L71
https://www.gnu.org/software/emacs/manual/html_node/elisp/Face-Attributes.html

Screenshots (with Iosevka):
Screenshot from 2022-08-19 14-12-59
image

Note that these screenshots are upscaled 200% from hidpi screen, so legibility of thin was even worse in practice.


thomasf:
sounds resonable

PR #441, opened 2023-08-04
fix compiler error

arteen1000:
Made changes to remove compiler error:

```
⛔ Warning (comp): solarized.el:252:2: Warning: docstring has wrong usage of unescaped single quotes (use \= or different quoting)
```

More information here:

https://emacs.stackexchange.com/questions/73047/emacs-29-docstring-single-quote-escaping-rules-compiler-level-event
https://www.gnu.org/software/emacs/manual/html_node/elisp/Keys-in-Documentation

I am not sure of the necessity in the case of the KEYSEQ.


thomasf:
thanks

PR #442, opened 2023-08-06
Add corfu face

rynffoll:
Added faces for Corfu package

## Screenshots

_Before_
image

_After_
image

-----------------

- [x] You've updated the readme (if adding/changing configuration options)

Thanks!

### DIFF
--- a/solarized-faces.el
+++ b/solarized-faces.el
@@ -433,6 +433,11 @@
      `(company-tooltip-search ((,class (:foreground ,magenta))))
      `(company-tooltip-search-selection ((,class (:foreground ,magenta :weight bold))))
      `(company-tooltip-selection ((,class (:weight bold))))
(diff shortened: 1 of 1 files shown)


rynffoll:
Example:

_Before_
image


_After_
image

thomasf:
I'm not wild about the currently selected match having less of a highlight than the unmatched ones.

This is whatever mode lsp+company mode uses for prompts in my config looks right now and the current selection is only differentiated by bold foreground

 image

rynffoll:
I guess I could try using `:inverse-video`, how about that?

```emacs-lisp
     `(corfu-current ((,class (:inherit corfu-default :weight bold :inverse-video t))))
```

It would look like this
image

thomasf:
Let's try that!

PR #451, opened 2024-04-16
Highlight font-lock-number-face as highlight-numbers-number

imustafin:
We already support `highlight-numbers` (by highlighting `highlight-numbers-number`). This PR uses the same font for `font-lock` (highlighting `font-lock-number-face`).

This fixes the inconsistency if you use both `highlight-numbers` and something which uses `font-lock`. For example, `ruby-mode` and `ruby-ts-mode`. For example, notice how the numbers are not highlighted only in the top-left cell in the following table:

<table>
<tr><th>Before</th><th>After</th></tr>
<tr>
<td>

image
</td>
<td>

image
</td>
</tr>
<tr>
<td>

image
</td>
<td>

image
</td>
</tr>
</table>
-----------------

- [x] You've updated the readme (if adding/changing configuration options)

Thanks!

### DIFF
--- a/solarized.el
+++ b/solarized.el
@@ -52,6 +52,13 @@ Related discussion: https://github.com/bbatsov/solarized-emacs/issues/158"
   :type 'boolean
   :group 'solarized)
 
+(defcustom solarized-highlight-numbers nil
+  "Highlight all numbers.
+Applies color to `font-lock-number-face' and `highlight-numbers' mode.
+Many tree-sitter based modes use `font-lock-number-face'."
+  :type 'boolean
+  :group 'solarized)
+
 (defcustom solarized-use-variable-pitch t
   "Use variable pitch face for some headings and titles."
   :type 'boolean
--- a/README.md
+++ b/README.md
@@ -106,6 +106,9 @@ elisp code:
 (setq solarized-height-plus-3 1.0)
 (setq solarized-height-plus-4 1.0)
 
+;; Highlight all numbers
+(setq solarized-highlight-numbers t)
+
 ```
 
 Note that these need to be set **before** `load-theme` is invoked for Solarized.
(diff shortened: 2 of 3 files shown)


thomasf:
This theme aims to be minimalist, I don't think anything that could enable different colors for numbers unless the user opts in to it by enabling a specific mode or customization is a good fit.

Having said that, I just took a quick look at a couple of programming modes and the font-lock-number-face wasn't applied to numbers in either one of them.

In any case, if you really think it's important, add a defcustom that defaults to being off for it.

thomasf:
This is a very similar kind of customization

```
(defcustom solarized-distinct-doc-face nil
  "Make `font-lock-doc-face' stand out more.
Related discussion: https://github.com/bbatsov/solarized-emacs/issues/158"
  :type 'boolean
  :group 'solarized)
```

bbatsov:
@thomasf Btw, this face is used by all modes using TreeSitter to my knowledge, as is the newer `font-lock-variable-usage-face`, etc. But whether numbers really need a separate face is a subject for debate and also something everyone can customize easily for themselves.

imustafin:
@thomasf thanks for the suggestion. I moved highlight behind an option.

Treesit-based modes often use font lock and set `font-lock-number-face`. This includes `ruby-ts-mode` and `js-ts-mode` to name a few.

This PR unifies color used for `highlight-numbers-number` (from `highlight-numbers` mode) and `font-lock-number-face`. With this PR both of them will be controlled using one option which is off by default.

I see one problem in my implementation. Those who had `highlight-numbers` installed will lose the highlight until they enable `solarized-highlight-numbers`. Is this a big problem? How can we improve this?

solarized.el
```diff
@@ -52,6 +52,13 @@ Related discussion: https://github.com/bbatsov/solarized-emacs/issues/158"
   :type 'boolean
   :group 'solarized)
 
+(defcustom solarized-highlight-numbers nil
```
bbatsov, line 55: I'd probably name this `solarized-font-lock-numbers`, so it's more in line with Emacs conventions.

PR #461, opened 2026-07-04
fix terminal history autosuggestions

cospectrum:
[Autosuggestions](https://github.com/zsh-users/zsh-history-substring-search) were indistinguishable from plain text.

Before:
Screenshot 2026-07-05 at 01 06 30
After:
Screenshot 2026-07-05 at 01 24 24


bbatsov:
Thanks!

